PR #38, opened 2019-08-08
CI: build on Windows

caelunshun:


caelunshun:
Well... time to not use OpenSSL.

PR #42, opened 2019-08-10
Rewrite networking code to use Tokio

### LINKED ISSUE #34: Worker threads panic when client is disconnected during the same poll cycle as a packet send request

### LINKED ISSUE #35: Switch to Tokio from Mio
This is the solution to #34 and #36.

### LINKED ISSUE #36: Mojang API requests in initial handler block the IO worker

caelunshun:
This PR includes a refactoring and partial rewrite of the networking code, based on [`tokio`](https://github.com/tokio-rs/tokio).

## Motivation
The current network code is a homebrew asynchronous IO event loop based on [`mio`](https://github.com/tokio-rs/mio). Numerous issues have appeared with this implementation, and the code is quite inefficient as well (three to five `HashMap` lookups when data is received or sent).

Tokio provides its own heavily optimized event loop, in addition to a much easier to use interface. Furthermore, it can extend the async IO capability beyond the TCP streams themselves; for example, the authentication request to the Mojang API in the initial handler can now be run asynchronously, which improves performance (somewhat).

Ultimately, the decision to switch to Tokio is based on performance and stability, which both are improved through the switch.

## Implications
* Server packet handling code will be unaffected.
* The new networking code uses `async`/`await`, which was stabilized a few weeks ago and will be released in Rust stable in early to mid November. In the meantime, compiling Feather will require Rust nightly.

Fixes #34 .
Resolves #35.
Resolves #36.


caelunshun:
Functionality is complete. Only thing left is to fix tests and clippy.

PR #64, opened 2019-08-19
Implement server list icon: (#37)

momothereal:
Implements the server list icon functionality by loading the file at `server-icon.png` and sending it in the server list response packet.

This implementation does not perform any checks on the PNG (format, size, etc.). The client will simply ignore the favicon if the image is invalid.

Protocol reference: https://wiki.vg/Server_List_Ping#Response

### DIFF
--- a/server/src/io/initialhandler.rs
+++ b/server/src/io/initialhandler.rs
@@ -117,7 +117,11 @@ pub struct InitialHandler {
 }
 
 impl InitialHandler {
-    pub fn new(config: Arc<Config>, player_count: Arc<PlayerCount>, server_icon: Arc<Option<String>>) -> Self {
+    pub fn new(
+        config: Arc<Config>,
+        player_count: Arc<PlayerCount>,
+        server_icon: Arc<Option<String>>,
+    ) -> Self {
         Self {
             action_queue: vec![],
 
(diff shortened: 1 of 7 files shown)


server/src/io/initialhandler.rs
```diff
@@ -205,6 +208,7 @@ fn handle_handshake(ih: &mut InitialHandler, packet: &Handshake) -> Result<(), E
 
 fn handle_request(ih: &mut InitialHandler, packet: &Request) -> Result<(), Error> {
     check_stage(ih, Stage::AwaitRequest, packet.ty())?;
+    let server_icon = (*ih.server_icon).clone().unwrap_or(String::new());
```
caelunshun: No need to dereference using `(*ih.server_icon) here—the compiler does that automatically.

server/src/main.rs
```diff
@@ -26,6 +26,7 @@ extern crate feather_codegen;
 extern crate bitflags;
 #[macro_use]
 extern crate feather_core;
+extern crate base64;
```
caelunshun: With Rust 2018, you don't have to do this anymore.

caelunshun:
Thanks for the help! It does look like the tests are failing to compile, though—the third parameter for initializing the IO manager in `testframework` is missing on line 44.

Other than that, this looks good.

server/src/io/initialhandler.rs
momothereal: I'm not too familiar with Arc, but if I try to use:

`ih.server_icon.clone().unwrap_or(String::new())`

I get an `error[E0507]: cannot move out of an Arc` at compile-time. De-referencing explicitly seemed to fix this, unless there is another way?

caelunshun: Yep, I guess you're right. The error happens because you're cloning the `Arc` and then unwrapping the `Option`, which consumes it. The alternative would be to use `Option::as_ref()`. but this is fine.

caelunshun:
Thanks!

PR #75, opened 2019-08-21
Add player join message

momothereal:


momothereal:
Curious about your opinion on fitting chat in the architecture. Do you figure there will be some utility functions to send chat messages easily, instead of broadcasting a packet all the time? Would it be a generic "JSON Chat Broadcast" event that is then broadcasted as a packet in a system?

caelunshun:
> Curious about your opinion on fitting chat in the architecture. Do you figure there will be some utility functions to send chat messages easily, instead of broadcasting a packet all the time? Would it be a generic "JSON Chat Broadcast" event that is then broadcasted as a packet in a system?

Yes, I would prefer to make the `PlayerChatEvent` more generic and extend it to all events involving chat messages. This wouldn't be so difficult—just change the event's name to `ChatSendEvent` and trigger it in the join event with the desired message.

In the future, the `ChatSendEvent` could also be extended to allow for custom JSON formatting and the like, but unfortunately I haven't implemented a library to do so yet.

momothereal:
Sounds good, I'll look into something like that tonight and implement it for this PR.

momothereal:
I changed the system for a more generic `ChatBroadcastEvent`. I'm not sure how I would write to the event queue in the `Util` struct however.

caelunshun:
Yes, this looks good. The event-based system will do for now.

If you wanted to use the `Util` type, you would maintain a `pending_chat_messages: SegQueue<String>` inside the struct, as is done with the `Spawner` type. Rather than writing to an event channel, `Util::broadcast_message` would push the message to the queue, and the `ChatBroadcastSystem` would flush the queue and send the messages.

As we won't need to send a lot of chat messages, doing so is probably unnecessary, though.

PR #80, opened 2019-08-23
Fix inventory indices - fixes #79

### LINKED ISSUE #79: Correctly handle inventory indexing
> I noticed that the handling of the Creative Item packet doesn't take into account the "raw slot" logic:
>
> https://github.com/caelunshun/feather/blob/d3d05d34eefe2047c374a66266e35715ebd9e2b1/server/src/player/inventory.rs#L184-L191
>
> Essentially, `packet.slot` won't match the "inventory slot" properly. For example, "Creative Slot #0" > is not the hotbar, but the helmet slot. This mapping depends on the "top view" and "bottom view" of the inventory (think of the slots in a crafting bench, vs. the slots in a survival inventory with the equipment).
>
> In Bukkit, there is this concept of "InventoryView" which has a convert-slot function to handle this.
>
> This is an issue because there is a desync as to which item is currently held by the player: using the "hotbar index" for the inventory slot will give the wrong item.

_Originally posted by @momothereal in https://github.com/caelunshun/feather/issues/46#issuecomment-524460222_

caelunshun:
This is a work-in-progress fix to #79, using an enum type containing variants for every possible slot index and a function to map variants of the enum to protocol indices for a given inventory type.

### DIFF
--- a/core/src/inventory.rs
+++ b/core/src/inventory.rs
@@ -4,39 +4,146 @@ use crate::item::Item;
 use smallvec::SmallVec;
 use std::cmp::min;
 
-pub type SlotIndex = usize;
+/// All possible slots in an inventory.
+///
+/// Some terms which are used here:
+/// * "Hotbar" refers to the hotbar.
+/// * "Main" refers to the 27 slots of the player
+/// inventory not contained within the hotbar,
+/// armor, and crafting slots.
+/// * "Crafting" refers to the crafting area of a player's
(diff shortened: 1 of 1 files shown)


caelunshun:
Yikes... implementing this macro is going to be painful.

caelunshun:
This PR is going to have to be rewritten, given the number of changes in the codebase since it was implemented.

Issue #72, opened 2019-08-20
Player data loading


caelunshun:
Add support for loading player data files (located in `world/playerdata/<uuid>.dat`.) This includes loading the player's position and inventory.

Related: #71

momothereal:
The `player.dat` file is now loaded (#96), but only for gamemode and last-known position & rotation. Inventory loading is not implemented yet, but it might depend on #79?

PR #96, opened 2019-08-30
Start work on player-data loading (#72); default gamemode

momothereal:
Start work on loading player NBT data when it exists. 

For the moment, it only loads the gamemode and last save pos, so it does not complete the scope of ticket #72. It should however be the groundwork for loading inventory data.


momothereal:
There are no tests yet because I'd like to get some feedback on the NBT-loading/component integration.

core/src/save/player_data.rs
```diff
@@ -1 +1,33 @@
...
+#[derive(Default, Debug, Clone, Serialize, Deserialize)]
+pub struct PlayerData {
+    #[serde(default = "default_gamemode")]
+    #[serde(rename = "playerGameType")]
+    pub gamemode: i32,
+}
+
+/// Set to a magic value when player data doesn't exist (first-time join)
+fn default_gamemode() -> i32 {
+    // FIXME: Doesn't seem to work for default (non-existing NBT files)
```
momothereal: I'm not sure if this is a limitation of the NBT deserialization library, but this seems like the way to do it according to the serde [docs](https://serde.rs/attr-default.html). Seems to default to `0` no matter what.

momothereal: Switched to using `None`, which makes it cleaner imo.

server/src/joinhandler.rs
```diff
@@ -167,8 +173,12 @@ impl<'a> System<'a> for JoinHandlerSystem {
                         continue;
                     }
 
-                    let spawn_block_pos =
-                        BlockPosition::new(level.spawn_x, level.spawn_y, level.spawn_z);
+                    let pos = positions.get(player).unwrap();
+                    let spawn_block_pos = BlockPosition::new(
```
caelunshun: This should use `Position::block_pos()` rather than a manual conversion.

core/src/lib.rs
```diff
@@ -53,6 +53,26 @@ impl Gamemode {
...
+        match id {
+            0 => Gamemode::Survival,
+            1 => Gamemode::Creative,
+            2 => Gamemode::Adventure,
+            3 => Gamemode::Spectator,
+            _ => Gamemode::Survival,
+        }
+    }
+
+    pub fn from_string(name: String) -> Self {
```
caelunshun: `name` should be passed as `&str`, since an owned value isn't needed here.

server/src/player/init.rs
```diff
@@ -43,21 +46,38 @@ impl<'a> System<'a> for PlayerInitSystem {
...
             };
             player_comps.insert(event.player, player_comp).unwrap();
 
-            let spawn_pos = position!(
-                f64::from(level.spawn_x),
-                f64::from(level.spawn_y),
-                f64::from(level.spawn_z)
-            );
+            let spawn_pos = match player_data.position {
+                Some(pos) => position!(pos[0], pos[1], pos[2]),
```
caelunshun: This will panic if the player data file doesn't contain 3 elements in the position array. A length check should be added before this.

server/src/player/init.rs
```diff
@@ -43,21 +46,38 @@ impl<'a> System<'a> for PlayerInitSystem {
             mut entity_types,
             mut metadata,
             level,
+            config,
         ) = data;
 
         // Run through events
         for event in events.read(&mut self.join_event_reader.as_mut().unwrap()) {
+            // Load player data
```
caelunshun: Rather than using `Option` in the player data struct and then using `Result::unwrap_or_default()`, I would recommend matching on the result from `load_player_data` like this (although this code won't compile):

```rust
let (position, gamemode) = match feather_core::player_data::load_player_data(uuid) {
    Ok(data) => (data.pos, data.gamemode),
    Err(e) => (position!(level.x, level.y, level.z), Gamemode::from_string(default_gamemode)),
};
```

server/src/joinhandler.rs
```diff
@@ -167,8 +173,8 @@ impl<'a> System<'a> for JoinHandlerSystem {
                         continue;
                     }
 
-                    let spawn_block_pos =
-                        BlockPosition::new(level.spawn_x, level.spawn_y, level.spawn_z);
+                    let pos = positions.get(player).unwrap();
+                    let spawn_block_pos = pos.current.block_pos();
 
                     let spawn_position = SpawnPosition::new(spawn_block_pos);
```
caelunshun: The `SpawnPosition` packet is used to send the world's spawn position, i.e. the location where a compass would point to. This change should be reverted.

momothereal: Ah, right. The PlayerPositionAndLookClientbound packet below will need to use different values then.

caelunshun:
Thanks!

PR #97, opened 2019-08-31
Load player inventory from save (#72)

### LINKED ISSUE #72: Player data loading

momothereal:
Closes #72.

### DIFF
--- a/core/src/save/player_data.rs
+++ b/core/src/save/player_data.rs
@@ -93,6 +93,7 @@ pub fn load_player_data(uuid: Uuid) -> Result<PlayerData, nbt::Error> {
 mod tests {
     use super::*;
     use crate::Gamemode;
+    use hashbrown::HashMap;
     use std::io::Cursor;
 
     #[test]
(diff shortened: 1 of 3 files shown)


core/src/save/player_data.rs
```diff
@@ -13,6 +15,46 @@ pub struct PlayerData {
...
+            crate::inventory::SLOT_HOTBAR_OFFSET + (self.slot as usize)
+        } else if self.slot == -106 {
+            // Offhand
+            crate::inventory::SLOT_OFFHAND as usize
+        } else if 100 <= self.slot && self.slot <= 103 {
+            // Equipment
+            (-self.slot + 108) as usize
+        } else {
+            // Rest of inventory
+            self.slot as usize
```
caelunshun: Consider verifying that the slot is valid here—invalid values (e.g. negative values, which would overflow when converted to unsigned) would cause a panic later on.

server/src/player/init.rs
```diff
@@ -100,7 +106,13 @@ impl<'a> System<'a> for PlayerInitSystem {
                 .insert(event.player, loaded_chunk_comp)
                 .unwrap();
 
-            let inventory_comp = InventoryComponent::new();
+            let mut inventory_comp = InventoryComponent::new();
+            for slot in inventory_slots {
+                let slot_index = slot.convert_index();
+                if slot_index.is_some() {
+                    inventory_comp.set_item_at(slot_index.unwrap(), slot.to_stack());
```
caelunshun: Unwrapping the slot index here will still cause a panic when the data file is invalid. Generally, panics should be avoided unless the error is critical—instead, a warning should be displayed and the slot skipped.

momothereal: I am checking if `slot_index.is_some()` before, so unwrapping should never occur if the index is invalid. I can add a warning as an else-statement if it's necessary.

> If the data-file is invalid

If that is the case, won't it be thrown when loading the file, and thus ignored by https://github.com/caelunshun/feather/blob/cb2d1e0a88e01ce0ef8a9806648f94f7c628f38a/server/src/player/init.rs#L67-L71

caelunshun: Ah, I missed that. Sorry.

When I said that the data file could be invalid, I meant that it could contain valid NBT data but the inventory indices would be invalid, which would cause a panic. I didn't notice the `is_some` check, which prevents the panic.

caelunshun: The more idiomatic syntax would be `if let Some(slot_index) = slot_index` rather than `is_some` and then `unwrap`, though.

momothereal: Right, the Clippy in CI is telling me about it, but my local install isn't for some reason. I'll make the change.

caelunshun:
Thanks again.

momothereal:
My pleasure!

Issue #24, opened 2019-07-28
World generation


caelunshun:
Ha... this is going to be difficult.

momothereal:
I think a first step would be to load the `generatorName` and `generatorOptions` fields from the `level.dat` file.

Then, supporting basic super-flat generation using the layers in `generatorOptions` to set blocks in un-generated chunks wouldn't be too hard. Afterwards, we'd need basic decorators/populators for biome features like trees as defined in the options' `structures`.

We'll need a perlin-noise library to support terrain generation in "default" world types, as well as implementing the biome grid/map layers. The logic for the latter could be based off [Glowstone](https://github.com/GlowstoneMC/Glowstone/tree/dev/src/main/java/net/glowstone/generator/biomegrid), which has near-parity with Vanilla (but only for 1.12).

Larger structures like villages, caves, and strongholds will probably be later additions and be tracked in separate tickets when the time comes.

At some point we'll also have to be able to generate a world from scratch (including the `level.dat`).

Does this plan seem fair to you, @caelunshun?

caelunshun:
Seems reasonable to me. Superflat generation does seem like a good way to start, given its simplicity.

I think generating a "bland" default world should be a good next step—just simple use of Perlin noise to generate hilly variations. This would then lay the groundwork for biomes, caves, and such.

https://github.com/jackmott/rust-simd-noise seems like a good noise library to use—its benchmarks show it to be over eight times faster than alternatives thanks to SIMD.

cheako:
I was thinking about looking for a save file.  Links and meaning a repo would help....   someone could generate saves for a handful of seeds and post them...  just make sure you inject some art into selecting the seeds.  A website or application that generates saves would be better.

I don't know much about MC other than I think I would like it, I just won't run MS applications.

hansihe:
I implemented the Minecraft worldgen in this crate a few years ago

https://github.com/hansihe/voxel_worldgen

It has most of the base features including biome support and cavegen. I don't really care that much about it anymore, so you have my permission to change the license/incorporate it in the project as long as I am credited.

Tracreed:
Great resource for how certain things in the latest Minecraft update is achieved. A lot of talk about world generation.
https://www.youtube.com/watch?v=TycBrFKEteU

PR #108, opened 2019-09-03
[WIP] Superflat layer generation (#24)

momothereal:
- [x] Load `generatorName` and `generatorOptions` from level
- [x] Hook into `chunkworker` loading to generate new chunks
- [x] Generate superflat chunks using the layers defined in the `generatorOptions`
- [x] Do generation asynchronously
- [ ] Add tests


caelunshun:
Thanks for the work on this.

For chunk generation, I would recommend a `WorldGenerator` trait:
```rust
pub trait WorldGenerator {
    /// Generates the chunk at the given position.
    fn generate_chunk(&self, position: ChunkPosition) -> Chunk;
```

The chunk worker would then have a `Box<dyn WorldGenerator>` field which it would invoke in a [Rayon task](https://docs.rs/rayon/1.2.0/rayon/fn.spawn.html) for generating chunks.

If you have any questions, feel free to ask.

momothereal:
@caelunshun Thanks, this is a bit ambitious for my Rust experience so I'm sure I'll have a few questions regarding rayon :)

server/src/chunkworker.rs
```diff
@@ -72,12 +75,18 @@ struct ChunkWorker {
...
+    world_generator: Arc<dyn WorldGenerator>,
 }
 
 /// Starts a chunk worker on a new thread.
 /// The returned channels can be used
 /// to communicate with the worker.
-pub fn start(world_dir: &str) -> (Sender<Request>, Receiver<Reply>) {
+pub fn start(
+    world_dir: &str,
+    world_gen: Arc<dyn WorldGenerator>,
```
momothereal: So I'm trying to figure out the correct way to pass a `WorldGenerator` instance all the way from the `main.rs` to the `chunkworker.rs`. I see that `chunklogic.rs` is setting-up chunkworker, but I don't know how I would access a component from there. Any pointers?

caelunshun: Yes, this is an annoying caveat of Specs. To initialize the `ChunkWorkerHandle` resource, Specs calls `ChunkWorkerHandle::default` when `Dispatcher::setup` is called. There's no way to access components from the `default` function.

There is a solution in the Specs book [here](https://slide-rs.github.io/specs/04_resources.html#default-for-resources). You should likely use the second solution, which is to use `ReadExpect<'a, ChunkWorkerHandle>` rather than `Read` and then insert the chunk worker handle into the world in `ChunkLoadSystem::setup`. Here, you can access any components and resources you need for the chunk worker.

momothereal: So if I understand correctly, using a `ReadExpect` removes the need for the Default trait on the `ChunkWorkerHandle`. However, the `chunkworker` is also initialized a bit further down: https://github.com/caelunshun/feather/blob/01d6d08fc1bcadc3788e375e1bb380ce6ffaed03/server/src/chunk_logic.rs#L83-L91

I'm still a bit unclear on how I can access a component from there. I can add a `Read<Arc<dyn WorldGenerator>>` to the `SystemData` type, but I don't know how to unpack it from the `setup` function (with only `&mut world` provided).

Perhaps this sort of component dependency is not a "correct" way of tackling this?

caelunshun: Yes, your understanding of `ReadExpect` is correct.

You can access resources from the world using `World::fetch` and `World::fetch_mut`. To get the level data, for example:
```rust
let level = world.fetch_mut::<LevelData>();
```

momothereal:
Got it to work, it generates the new chunks using the layers. Just gotta make it asynchronous and write some unit tests.

Also I noticed that the `Block` enum entries that have property maps, like `GrassBlock`, will return an empty result if a property is missing (note the first `?`):
https://github.com/caelunshun/feather/blob/fc6818582144e9f3902e0e85fc20b40aadec8942/blocks/src/blocks.rs#L5620

Which makes the `from_name_and_props` return a None  when no properties are provided. Since the LevelData's `generatorOptions` layers doesn't contain any block data, it would have to get the default property map for the ID, or replace the `?` with `.unwrap_or` and some default state for each property.

caelunshun:
> Also I noticed that the `Block` enum entries that have property maps, like `GrassBlock`, will return an empty result if a property is missing (note the first `?`):
> https://github.com/caelunshun/feather/blob/fc6818582144e9f3902e0e85fc20b40aadec8942/blocks/src/blocks.rs#L5620
> 
> Which makes the `from_name_and_props` return a None when no properties are provided. Since the LevelData's `generatorOptions` layers doesn't contain any block data, it would have to get the default property map for the ID, or replace the `?` with `.unwrap_or` and some default state for each property.

Yes, I guess `from_name_and_props` should default to the default state when no properties are provided. As a quick fix, I could just change the generated `?` to `.unwrap_or_default()`.

momothereal:
Test `test_worldgen_flat` will fail as long as the above issue with the property map is resolved.

caelunshun:
Right, I'll get to implementing that then.

momothereal:
Sounds good. Also, I'm not super satisfied with the design of my changes in chunkworker, it seems a bit hacky to use Option/None when loading the chunk is delegated to the rayon task. It's also pretty annoying to test. Let me know if you have some ideas.

caelunshun:
Well, you could pass the channel to `load_chunk` and make sending the reply a side-effect of the function... but this might not be ideal either.

Anyway, in the Tokio rewrite (#42), the chunk worker will be rewritten to use Tokio tasks for IO and Rayon tasks for generation, which will remove the need to return an `Option`. I think this will be fine as it is.

momothereal:
Alright, I'll leave it as is then.

caelunshun:
@momothereal I implemented `Block::from_name_and_default_props`.

core/src/save/level.rs
```diff
@@ -77,6 +85,74 @@ pub struct LevelVersion {
...
+                },
+            ],
+            biome: String::from("minecraft:plains"),
+        }
+    }
+}
+
+#[derive(Default, Debug, Clone, Serialize, Deserialize)]
+pub struct SuperflatLayer {
+    pub block: String,
```
caelunshun: Should this not just be `Block`? I don't think we need to know the string value, just the block type.

server/src/chunkworker.rs
```diff
@@ -145,21 +155,42 @@ fn load_chunk(worker: &mut ChunkWorker, pos: ChunkPosition) -> Reply {
...
     let result = handle.load_chunk(pos);
 
     match result {
-        Ok(chunk) => (pos, Ok(chunk)),
+        Ok(chunk) => Some((pos, Ok(chunk))),
         Err(e) => match e {
-            region::Error::ChunkNotExist => (pos, Err(Error::ChunkNotExist)),
-            err => (pos, Err(Error::LoadError(err))),
+            region::Error::ChunkNotExist => {
+                let sender = Arc::clone(sender);
```
caelunshun: Rather than wrapping the channel in an `Arc` and cloning it, you can just clone the channel directly. https://docs.rs/crossbeam/0.7.2/crossbeam/channel/index.html#sharing-channels

core/src/save/level.rs
momothereal: I could be wrong, but that would mean implementing Serialize and Deserialize traits manually, since `Block` doesn't map from/to the NBT string type.

caelunshun: I think you can use https://serde.rs/field-attrs.html#serialize_with.

momothereal: Deserialization seems feasible, but implementing Serialization isn't feasible at the moment since there is no Block-to-identifier matching in the `block.rs` generated file. Implementing Serialize is required to implement Deserialize / use `deserialize_with` macro.

caelunshun: Ah, in that case, you can just leave it as is. Perhaps add a TODO so that we can change this once those mappings are generated.

caelunshun:
`test_spawn_item` is failing because `ChunkCrossSystem` depends on `ChunkWorkerHandle` but the handle is not initialized. Try inserting a chunk worker handle in `testframework::Builder::build()`.

caelunshun:
Ready to merge?

caelunshun:
Ah, hold on, there's still an issue with this. When a region file containing the chunk is missing, the chunk is not generated and instead an error is returned:
```
2019-09-06 19:27:45,806 WARN  [feather_server::chunk_logic] Failed to load chunk at ChunkPosition { x: 67, z: 21 }: Error loading chunk: No such file or directory (os error 2)
```

Consider generating the chunk if `Err` is returned here: https://github.com/caelunshun/feather/blob/aab3ebe0deb7a3341e7b0f1acc2f9a6e474a81b9/server/src/chunkworker.rs#L135

momothereal:
Ready to merge, unless there's something I can do about coverage.

caelunshun:
Thanks!

PR #114, opened 2019-09-04
Update specs dependency

flohag:
The current specs dependency causes the build to fail, because a commit seems to be missing.
Since the `storage-event-control` feature has been merged into the upstream repo we can just use this instead.


caelunshun:
Thanks!

PR #120, opened 2019-09-07
Basic Overworld world generation

caelunshun:
The aim is to implement a basic generator with Perlin noise and biome support. Caves, trees, etc. will be added in a later PR.

- [x] Lay groundwork for world generation with a `ComposableGenerator` based on http://cuberite.xoft.cz/docs/Generator.html
- [x] Implement a basic height map generator
- [x] Implement a basic composition generator
- [x] Implement a biome grid generator
- [x] Generate composition based on biomes
- [x] Implement a density map generator using 3D noise instead of 2D
- [x] Generate height based on biomes
- [x] Interpolate between biome height values to prevent sharp cutoffs
- [x] Implement basic "finishers," e.g. shrubs, grass, snow
- [x] Add tests where possible


caelunshun:
First steps!
failed worldgen

caelunshun:
Height map generator is a bit messed up at the moment. Hope to figure out what's causing this problem.

NiLSPACE:
In Cuberite we moved away from heightmap generators in favor of shape generators https://github.com/cuberite/cuberite/pull/1604. Basically, a shape represents if a block should be empty (air) or filled. The reason for this is because it was much harder to create proper overhangs if we're generating heights.

madmaxoft:
Feel free to contact me with questions about terrain generation (I'm the author of most of the Cuberite Composable generator docs you mentioned).

caelunshun:
Yes, that's what I plan to switch to—the height map generator is just temporary. The issue, though, is that the noise function doesn't appear to produce smooth results, and there are visible lines between chunks. So switching to a density map will still cause issues.

This image is a visualization of what the noise function outputs when called for generated in chunks:
output

See jackmott/rust-simd-noise#9.

caelunshun:
@madmaxof so I have a question about the biome height interpolation method mentioned in the Cuberite docs. I can see how you could interpolate between heights by averaging when using a 2D height map, but the same doesn't apply when using 3D noise to generate a density map rather than a height map. What's your method for doing this?

I see that Glowstone uses a 2D height noise with added 3D detail and roughness noises to add overhangs and such. They then use the 2D height noise to calculate interpolation between biomes. Is this a method you would recommend?

madmaxoft:
We use a bit different method.
For the base, we have two 3D noises that are used for the density, another low-frequency 3D noise is used to switch between them (this allows for sudden "walls", overhangs etc.) 
Then biomes averaged across 9 * 9 columns are used to select two "parameters" for each point, the ideal height and the ideal amplitude of the terrain. Each biome has those two hard-coded. The ideal height and the amplitude are then turned into a vertical "gradient" (height -> gradient midpoint, amplitude -> gradient slope) which is added to the noise. The gradient is slightly modified to account for some shaping - at the top of the chunk, the gradient makes the terrain disappear faster, and underground the gradient is 4 times "slopier" than above ground to make the ground have fewer holes. Finally, yet another 2D noise, added to each Y column, is used to modify the ground height.

Of course calculating all this would be too much for the CPU, so these noises are all linearly upscaled by a factor of 4 (so that only a 5 * 33 * 5 values from each are needed for a single chunk).

Here's the C++ code:
https://github.com/cuberite/cuberite/blob/7678d5e6ed6fcc9361416ef41c43fa09a1d49f6f/src/Generating/Noise3DGenerator.cpp#L585-L772

caelunshun:
I plan to merge this PR within the next day and make incremental improvements to worldgen over time. @momothereal any thoughts?

momothereal:
Just tested the PR. Very impressive work!

Sounds good to me

Issue #101, opened 2019-08-31
Handle arrow shooting


caelunshun:
When a `PlayerDigging` packet is received with status `ShootArrow`, spawn an arrow with appropriate velocity.

This will be a nice way to show off the new physics system (hopefully).

PR #121, opened 2019-09-08
Handle arrow shooting (#101)

momothereal:
#101
- [x] Handle digging status for using bow
- [x] Find arrow stack and consume in Survival/Adventure
- [x] Submit event for shooting arrow with given position
- [x] Implement Arrow entity
- [x] Spawn Arrow entity with velocity from player's rotation
- [x] Add tests


caelunshun:
For spawning the arrow, I recommend adding a method to `util::Spawner` to lazily spawn it, as is done with items. This will avoid the need for tons of component accesses in the player digging system.

momothereal:
> For spawning the arrow, I recommend adding a method to `util::Spawner` to lazily spawn it, as is done with items. This will avoid the need for tons of component accesses in the player digging system.

Yes, I was thinking of making a `PlayerShootBowEvent`, with a `PlayerShootBowSystem` listening for it. The system would then spawn the arrow using a function in `Spawner` similar to items. Does that seem right?

caelunshun:
Yes, that seems right. You might want to consider how you could generalize `PlayerShootBowEvent` to support other arrow shooting causes, such as dispensers.

caelunshun:
What's with the huge velocity for arrows?

caelunshun:
Sorry, accidental close

momothereal:
So I'm not too sure, but it seems like the scale of velocity is very different for items and arrows. I packet-logged Spigot and saw velocity values in the thousands.

According to https://wiki.vg/Protocol#Entity_Velocity, it should be pretty normal to get very high velocity values since it's measured in 1/8000th of blocks per tick. Why the values are so much smaller for items is unclear to me.

caelunshun:
The velocity value on Feather is in blocks per tick. The `protocol_velocity` function converts it to the units used for packets, which is what you're observing.

momothereal:
The thing is `util::protocol_velocity` is performing the opposite conversion: dividing by 8000 is converting the protocol format to blocks-per-tick.

caelunshun:
Ah... so that `protoocl_velocity` has always been incorrect, then. It should be multiplying by 8000. Sorry about that.

(Now I know what's causing entities to jitter on the client...)

momothereal:
I will merge develop into this branch and adjust the velocity.

momothereal:
Ready for review/merge.

server/src/entity/arrow.rs
```diff
@@ -0,0 +1,65 @@
...
+use crate::entity::NamedComponent;
+use crate::player::PLAYER_EYE_HEIGHT;
+use crate::util::Util;
+
+/// Component for arrow entities.
+#[derive(Default)]
+pub struct ArrowComponent;
+
+impl Component for ArrowComponent {
+    type Storage = VecStorage<Self>;
```
caelunshun: For marker components (those with zero size), `NullStorage` should be used.

https://slide-rs.github.io/specs/05_storages.html

server/src/player/digging.rs
```diff
@@ -236,6 +249,112 @@ fn handle_drop_item_stack(
...
+
+    shoot_arrow_events.single_write(ShootArrowEvent {
+        shooter: Some(entity),
+        position,
+        arrow_type,
+        critical: false, // TODO: Determine critical based on how long bow was pulled back
+    });
+}
+
+fn find_arrow(inventory: InventoryComponent) -> Option<(SlotIndex, ItemStack)> {
```
caelunshun: Passing a reference to `InventoryComponent` should suffice here. No need to clone the entire inventory, just clone the item stacks you retrieve.

caelunshun:
Also, physics aren't being applied to arrows on the server side because they don't have a bounding box (`EntityPhysicsSystem` joins over entities with bounding boxes). This can be observed by shooting an arrow and then rejoining; the arrow will be back to where it was shot from.

momothereal:
> Also, physics aren't being applied to arrows on the server side because they don't have a bounding box (`EntityPhysicsSystem` joins over entities with bounding boxes). This can be observed by shooting an arrow and then rejoining; the arrow will be back to where it was shot from.

I added a bounding box and made the arrow spawn 1.5 blocks in front of the player. However, it seems like the arrow breaks off its course after a few blocks and falls down. Seems like something is off with the physics impl?

caelunshun:
Yes, `physics::entity::drag_force` returns 0 by default, so the velocity gets set to 0. `gravitational_accelration` should return -0.05 and `drag_force` 0.99 for arrows. I tried this, and it solved the problem.

momothereal:
Cool. Is there a value set for the "gliding" effect when the entity is blocked? Arrows should do a full-stop when colliding with a block.

caelunshun:
That's not implemented yet, but you could just set `slip_multiplier` to 0.0 for arrows here. https://github.com/caelunshun/feather/blob/d9f973899e74f7816530e50f451da5b7592521a6/server/src/physics/entity.rs#L159

That won't work when the arrow hits a wall—only when it impacts the ground. A more robust solution would be to add an `EntityCollideEvent` and set arrows' velocities to 0 when the event is triggered.

caelunshun:
Thanks!

PR #123, opened 2019-09-08
Set biome from generatorOptions for flat chunks

momothereal:
Reads the Biome from SuperflatGeneratorOptions and sets it in the new chunk.


caelunshun:
Thanks!

Issue #102, opened 2019-08-31
Load entities in chunk data


momothereal:
Load the entities saved in the [ChunkLevel](https://github.com/caelunshun/feather/blob/97ee3dc57c581aa9e9b010b355fa9d79b8a92d3e/core/src/save/region.rs#L35) NBT compound. See [Entity Format](https://minecraft.gamepedia.com/Chunk_format#Entity_format) for reference.

Similar to #72. Implementing each entity (including broadcasting to the player) will be on a per-entity basis, similar to the `Item` and `Player` entities, and not in the scope of this ticket.

caelunshun:
One issue here is that the `SystemData` for whatever system is responsible for loading entities will be huge, since it will need access to practically every component storage. We might want to consider using `specs::LazyUpdate` to initialize entities so as to keep the code cleaner.

PR #125, opened 2019-09-10
Load item and arrow entities from chunk save (#102)

momothereal:
#102 
- [x] Load item NBT
- [x] Load arrow NBT
- [x] Spawn item and arrow into world
- [x] Add tests

### DIFF
--- a/core/src/save/entity.rs
+++ b/core/src/save/entity.rs
@@ -1,3 +1,5 @@
+use crate::Position;
+
 #[derive(Clone, Serialize, Deserialize, Debug)]
 #[serde(tag = "id")]
 pub enum EntityData {
@@ -22,11 +24,29 @@ pub struct BaseEntityData {
     pub velocity: Vec<f64>,
 }
 
+impl BaseEntityData {
+    /// Reads the position and rotation fields. If the fields are invalid, None is returned.
+    pub fn read_position(self: &BaseEntityData) -> Option<Position> {
+        if self.position.len() == 3 && self.rotation.len() == 2 {
+            Some(Position {
+                x: self.position[0],
+                y: self.position[1],
+                z: self.position[2],
+                yaw: self.rotation[0],
+                pitch: self.rotation[1],
+                on_ground: true,
+            })
+        } else {
+            None
+        }
+    }
+}
+
 /// Represents a single item, without slot information.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ItemData {
     #[serde(rename = "Count")]
-    pub count: i8,
+    pub count: u8,
     #[serde(rename = "id")]
     pub item: String,
 }
--- a/core/src/save/entity.rs
+++ b/core/src/save/entity.rs
@@ -1,0 +1,44 @@
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(tag = "id")]
+pub enum EntityData {
+    #[serde(rename = "minecraft:item")]
+    Item(ItemEntityData),
+
+    /// Fallback type for unknown entities
+    #[serde(other)]
+    Unknown,
+}
+
+/// Common entity tags.
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct BaseEntityData {
+    #[serde(rename = "Pos")]
+    pub position: Vec<f64>,
+    #[serde(rename = "Motion")]
+    pub velocity: Vec<f64>,
+}
+
+/// Represents a single item, without slot information.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ItemData {
+    #[serde(rename = "Count")]
+    pub count: i8,
+    #[serde(rename = "id")]
+    pub item: String,
+}
+
+/// Data for an Item entity (`minecraft:item`).
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct ItemEntityData {
+    // Inherit base entity data
+    #[serde(flatten)]
+    pub entity: BaseEntityData,
+
+    // Item-specific tags
+    #[serde(rename = "Age")]
+    pub age: i16,
+    #[serde(rename = "PickupDelay")]
+    pub pickup_delay: u8,
+    #[serde(rename = "Item")]
+    pub item: ItemData,
+}
--- a/core/src/save/entity.rs
+++ b/core/src/save/entity.rs
@@ -1,4 +1,4 @@
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Clone, Serialize, Deserialize, Debug)]
 #[serde(tag = "id")]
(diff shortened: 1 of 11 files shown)


core/src/save/region.rs
```diff
@@ -175,6 +178,21 @@ impl RegionHandle {
...
 
+        for entity in &level.entities {
+            match entity {
+                EntityData::Item(item_data) => {
+                    debug!("Found an item entity: {:?}", item_data);
+                    // TODO: Create item entity from data
+                }
+                EntityData::Unknown => {
+                    trace!(
+                        "Chunk {:?} contains an unknown entity type",
```
momothereal: It's unfortunate that serde's `#[serde(other)]` attribute restricts the Unknown entry to being a unit variant, since we can't fetch the stored type of the unknown entity.

caelunshun: This should be fine, I think. End users won't care anyway, and eventually all entity types will be implemented.

caelunshun:
What's your idea for spawning the loaded entities?

momothereal:
I'm not decided yet, but here's what I'm thinking:

Since level deserialization is in `feather-core`, I figure the entities would be stored in the `Chunk` struct. When `ChunkLoadSystem` (in `feather-server`) receives a `Reply` with a loaded chunk, it fires a `ChunkLoadEvent`. There could be another system in `feather-server` that listens on `ChunkLoadEvent` and spawns the entities depending on their data.

However, consider the case where an an entity is stored in Chunk `A`. It is added to the world when `ChunkLoadEvent` is fired for Chunk `A`. The entity then crosses to Chunk `B`. Chunk `A` is later unloaded, and then loaded again. The entity would then be duplicated in the world. Not too sure how to prevent that.

caelunshun:
> I'm not decided yet, but here's what I'm thinking:
> 
> Since level deserialization is in `feather-core`, I figure the entities would be stored in the `Chunk` struct. When `ChunkLoadSystem` (in `feather-server`) receives a `Reply` with a loaded chunk, it fires a `ChunkLoadEvent`. There could be another system in `feather-server` that listens on `ChunkLoadEvent` and spawns the entities depending on their data.
> 
> However, consider the case where an an entity is stored in Chunk `A`. It is added to the world when `ChunkLoadEvent` is fired for Chunk `A`. The entity then crosses to Chunk `B`. Chunk `A` is later unloaded, and then loaded again. The entity would then be duplicated in the world. Not too sure how to prevent that.

Storing entities in `Chunk` isn't ideal in my opinion, since they aren't needed past when the chunk is loaded (all entities are stored in the `World`). It would probably be preferable to add a `Vec<EntityData>` field to `chunkworker::Reply`.

As for the entity duplication issue, I don't think there's a clean way around this until world persistence is implemented. I plan to do this within the next few weeks, so the issue shouldn't be present for that long.

momothereal:
Would it make sense to add the `Vec<EntityData>` to `ChunkLoadEvent`, or firing a separate event?

caelunshun:
A separate system makes the most sense in this case, yes.

momothereal:
Yeah that's what I'm doing atm, though it's fetching the `Util` resource lazily. Would that still be beneficial in a separate system, or should I just use `Read<'a, Util>` directly?

caelunshun:
Still a separate system, I think, because the match expression will grow very long.

momothereal:
Oh yeah I'm for sure separating it, I was just wondering if there'd be a purpose in using `LazyUpdate` in that new system, instead of referencing `Util` directly. I guess I'm not clear on what `LazyUpdate` actually does.

caelunshun:
Oh, sorry for misunderstanding.

`LazyUpdate::exec` causes the provided closure to be execute lazily, i.e. at the end of the tick. Its benefit is that you can access any resource or component from the world without specifying system data, but you also lose concurrency by doing this.

If you use the separate system, there shouldn't be a need to use it. The main motivation for utilizing it is if you need access to tons of components which would be unclean to write in a `SystemData`. (In the issue, I wrote that LazyUpdate should be used, but this was before `util::spawn_xx` was implemented.)

momothereal:
Hmm so if I add `Vec<EntityData>` to `ChunkLoadEvent`, I get an error since `Vec` doesn't implement the Copy trait. Does that mean I can't pass a Vec in an event?

caelunshun:
You should just be able to remove the `Copy` derive on `ChunkLoadEvent` (unless there's some reason it's needed).

core/src/save/entity.rs
```diff
@@ -0,0 +1,80 @@
...
+/// Data for an Arrow entity (`minecraft:arrow`).
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct ArrowEntityData {
+    // Inherit base entity data
+    #[serde(flatten)]
+    pub entity: BaseEntityData,
+
+    // Arrow-specific tags
+    #[serde(rename = "crit")]
+    pub critical: u8,
```
momothereal: For some reason, it's not able to deserialize `TAG_Byte` as `bool` in those structs. It logs an error saying `Region file contains invalid NBT: invalid type: integer 0, expected a boolean`. I'm not sure why since `LevelData` contains multiple `bool` fields that load just fine.

caelunshun: Odd, maybe the value changed to a TAG_Byte in 1.13? I know the map format documentation on both the Minecraft wiki and wiki.vg is outdated. I'll inspect the NBT when I get home.

caelunshun: Hmm... well, it's probably a problem with hematite-nbt. We could file an issue there, but I'm not sure of what the cause could be...

momothereal: I identified a reproducible test case and submitted an issue there: https://github.com/PistonDevelopers/hematite_nbt/issues/43

caelunshun: Great. Would you like to merge now or after that issue is resolved?

momothereal: I'm going to add some more tests and documentation to this PR first, but I don't think we should block on that issue. I'll add a todo to change that crit field to bool.

momothereal:
Ready for review

server/src/entity/chunk.rs
```diff
@@ -155,13 +159,79 @@ impl<'a> System<'a> for ChunkEntityUpdateSystem {
...
+                    // TODO: Spawn remaining entity types here.
+                    EntityData::Unknown => {
+                        trace!("Chunk {:?} contains an unknown entity type", event.pos);
+                    }
+                }
+            }
+        }
+    }
+
+    fn setup(&mut self, world: &mut World) {
```
caelunshun: There's a macro for implementing the setup function for event readers. `setup_impl!(reader)` should do the trick here.

caelunshun:
Thanks!

PR #127, opened 2019-09-12
Fix der encoding.

cheako:
This changes the data sent into rsa-der to be signed data, as expected by that crate.  The issue is at rsa-der/src/lib.rs#L59 where unsigned data is read as signed, everything after that is operating as it's supposed to.

Other options:
  Change rsa-dir to internally read in the data as unsigned, currently the data is read as signed: https://github.com/caelunshun/rsa-der/blob/8c17a00b3b81887c0e25b5cc5ca442754ace14c9/src/lib.rs#L59
  There is a patch for that: https://github.com/caelunshun/rsa-der/files/3599335/0001-This-is-a-minor-change-to-make-sure-the-numbers-load.patch.txt

In my opinion both are equivalent.  This pull has the effect of not altering the rsa-der API, but overall I think the patch is less intrusive.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -676,6 +676,7 @@ dependencies = [
  "nalgebra 0.18.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "nalgebra-glm 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "ncollide3d 0.20.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


server/.gitignore
```diff
@@ -0,0 +1,3 @@
+/world
```
caelunshun: If you want to ignore stuff that's only relevant for you, please put it in .git/info/exclude rather than .gitignore so that it isn't committed.

caelunshun:
Can you merge develop into this branch? We typically use develop for all development; master is only for the most recent stable release.

caelunshun:
Thanks.

Issue #104, opened 2019-09-01
Propagate block updates to adjacent blocks


caelunshun:
When a block update happens, adjacent blocks need to be updated accordingly. For example, sand and gravel should be turned into falling block entities, and grass, flowers, etc, should be broken when the block below them is destroyed.

A possible implementation of this:
* A registry is created as a Specs resource, containing the set of all blocks which require notification upon adjacent block updates.
* A `BlockUpdatePropagateSystem` listens to `BlockUpdateEvent`s and triggers a `BlockNotifyEvent` for each adjacent block type which exists in the registry.
* Separate systems are created for each category of block updates which need to be made. For example, a `FallingBlockCreationSystem` would listen to `BlockNotifyEvent`s, check if the block type is sand or gravel, and then perform the necessary actions.

Note that falling blocks are separate issues and are only used as examples here.

PR #129, opened 2019-09-13
Block update propagation and falling blocks

momothereal:
#104


server/src/entity/falling_block.rs
```diff
@@ -0,0 +1,19 @@
...
+impl Default for FallingBlockComponent {
+    fn default() -> Self {
+        FallingBlockComponent {
+            block: Block::Stone,
+        }
+    }
+}
+
+impl Component for FallingBlockComponent {
+    type Storage = VecStorage<Self>;
```
caelunshun: Probably use DenseVecStorage rather than VecStorage here, since most entities won't be falling blocks. VecStorage would waste memory in this case.

momothereal: Should `ItemComponent` be changed too, then?

caelunshun: Yes, I guess that's probably true. Not sure what I was thinking when I chose that storage, sorry for misleading.

server/src/blocks/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+                    for z in -1..=1 {
+                        let mut adjacent = event.pos;
+                        adjacent.x += x;
+                        adjacent.y += y;
+                        adjacent.z += z;
+                        let block: Option<Block> = chunk_map.block_at(adjacent);
+                        if let Some(block) = block {
+                            match block {
+                                Block::Sand | Block::RedSand | Block::Gravel => {
+                                    notify_events.push(BlockNotifyEvent {
```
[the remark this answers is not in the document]
caelunshun: I see, that makes sense. This list could grow pretty long, so consider a `lazy_static` `HashSet` containing all these blocks.

momothereal:
Ready for review

caelunshun:
Thanks, looks good!

PR #137, opened 2019-09-14
Break non-solid blocks when block underneath is broken

caelunshun:
#135

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -467,7 +467,7 @@ dependencies = [
 
 [[package]]
 name = "derivative"
-version = "1.0.2"
+version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "proc-macro2 0.4.30 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


momothereal:
I think the `non-solid` terminology is inconsistent here. Blocks like Slabs are supposed to be non-solid, since they have behavior like falling-blocks breaking when falling on them.

"Attached" blocks should be a separate group.

caelunshun:
> I think the `non-solid` terminology is inconsistent here. Blocks like Slabs are supposed to be non-solid, since they have behavior like falling-blocks breaking when falling on them. Cobwebs are also non-solid, but they are not attached.
> 
> "Attached" blocks should be a separate group.

Makes sense. I was generally under the impression that non-solid blocks are blocks that entities ignore in physics, while slabs still have bounding boxes and thus are non-solid. I guess blocks should be split into three groups, then, `non-solid` for those that are not complete blocks, `attachable` for those that require supporting blocks, and `empty` for those that do not affect physics.

momothereal:
In my opinion, it would be unwise to separate block types into larger "categories" based on their behavior, at least for now. There's a lot of edge-cases, like how slabs behave differently if they are "top" or "bottom" slabs.

For `attached` blocks, I think it's fine. Note that which adjacent block is supporting the block depends on the block's data, e.g. a wall torch facing north would break if its southern-adjacent block breaks.

I just think the terms `solid` and `empty` have a lot of overlapping definitions (solid in terms of physics? liquids? lighting?). `attached` is a good level of category, and another could be `pass_through` (no physics), another `opaque` for lighting, etc.

OvermindDL1:
For note, 'Attached' in MC, at least in the majority of the modding API's references if a block is 'attached' to another block, like a torch on a wall, and it also often defines 'where' on the block it is attached, generally Center, Top Edge, Bottom Edge, Left Edge, Right Edge, and each of the four corner edges are defined, which something potentially 'attached' to multiples of them, a set of 16 bits I always thought would have been most useful for that.  In addition, when something is attached and the attachment point changes to no longer match the bitmask (AND is 0) then they don't necessarily always pop up off, but potentially they can 'shift' to another attachment point on another face, which is determined by the block.  The vanilla torch does this but it is common in mods as well.

caelunshun:
With the Legion rewrite, this will have to be redone. I would also like to implement the block functionality described above.

PR #138, opened 2019-09-15
[WIP] Tree generation

caelunshun:
- [x] Implement a `Schematic` type for storing structures and writing parts of them which overlap with a chunk to the chunk
- [ ] Implement tree schematic generators
- [ ] Implement a tree finisher, with tree types/density based on biomes

### DIFF
--- a/server/src/worldgen/schematic.rs
+++ b/server/src/worldgen/schematic.rs
@@ -1,0 +1,226 @@
+use feather_blocks::Block;
+use feather_core::{Chunk, ChunkPosition};
+
+/// A 3D schematic of some structure, such as a tree or building.
+///
+/// Schematics are not resizable and have static dimensions.
+///
+/// # Notes
+/// * Positions in any parameter passed to a method here are
+/// local to the schematic. (0, 0, 0) corresponds to the center
+/// of the schematic, which is set in `new`. Negative indices
+/// are allowed.
+#[derive(Debug, Clone)]
+pub struct Schematic {
+    /// 3D grid of blocks. Length of this vector
+    /// is always length * width * height.
+    blocks: Vec<Block>,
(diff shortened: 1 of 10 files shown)


caelunshun:
I noticed a severe performance impact from adding this, and noticed a rather peculiar cause when investigating:
Screenshot from 2019-09-14 23-18-51

`bitvec` is for some reason the main overhead for worldgen, which seems somewhat odd...

server/src/worldgen/schematic.rs
```diff
@@ -0,0 +1,268 @@
...
+pub struct SchematicBuilder {
+    length: usize,
+    height: usize,
+    width: usize,
+    center_x: usize,
+    center_y: usize,
+    center_z: usize,
+}
+
+impl SchematicBuilder {
```
Unrealrussian, line 177: Wouldn't it be better to collapse with_dimensions into new. I don't see a case where dimensions 0,0,0 would be used.

server/src/worldgen/schematic.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn with_dimensions(mut self, length: usize, height: usize, width: usize) -> Self {
```
Unrealrussian, line 182: Wouldn't it be better to collapse with_dimensions into new. I don't see a case where dimensions 0,0,0 would be used.

server/src/worldgen/schematic.rs
Unrealrussian, line 177: Also a function where you would fill a region of blocks with a specific blocktype would be a great helper function. It is a repeated pattern in future Schematics, where you would fill a specific part (like walls, floors, treeleaves, etc.) in the schematics.

PR #140, opened 2019-09-16
Fix test_basic_composition_for_column (bitvec)

momothereal:


caelunshun:
Mmm, keep forgetting to update tests. Thanks.

PR #141, opened 2019-09-18
Create world level.dat when it doesn't exist

momothereal:
Creates a world if the `world` directory doesn't exist.

For now it will only write the `level.dat` file when creating it for the first time.

I added a config entry to change the world directory feather looks for (defaults to `world`). There's also options for `seed` and `generator`, only used when a new world is being created.

### DIFF
--- a/server/src/chunk_logic.rs
+++ b/server/src/chunk_logic.rs
@@ -86,8 +86,8 @@ impl<'a> System<'a> for ChunkLoadSystem {
         use specs::prelude::SystemData;
 
         let generator = world.fetch_mut::<Arc<dyn WorldGenerator>>().clone();
-        let config = world.fetch_mut::<Arc<Config>>().clone();
-        let world_dir = Path::new(&config.world.name);
+        let world_name = &world.fetch_mut::<Arc<Config>>().world.name.clone();
+        let world_dir = Path::new(world_name);
 
         info!("Starting chunk worker thread");
         let (sender, receiver) = chunkworker::start(world_dir, generator);
--- a/server/src/chunk_logic.rs
+++ b/server/src/chunk_logic.rs
@@ -13,6 +13,7 @@ use feather_core::world::{ChunkMap, ChunkPosition};
 
 use rayon::prelude::*;
 
+use crate::config::Config;
 use crate::entity::EntityDestroyEvent;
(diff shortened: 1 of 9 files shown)


server/src/chunk_logic.rs
```diff
@@ -84,9 +86,11 @@ impl<'a> System<'a> for ChunkLoadSystem {
         use specs::prelude::SystemData;
 
         let generator = world.fetch_mut::<Arc<dyn WorldGenerator>>().clone();
+        let config = world.fetch_mut::<Arc<Config>>().clone();
```
caelunshun: Rather than fetching and cloning the `Arc`, I would recommend just cloning `config.world.name`.

server/src/main.rs
```diff
@@ -112,8 +116,21 @@ fn main() {
         Arc::clone(&server_icon),
     );
 
-    info!("Loading level.dat");
-    let level = load_level().unwrap_or_else(|e| {
+    let world_name = &config.world.name;
+    let world_dir = Path::new(world_name.as_str());
+    let level_file = &world_dir.join("level.dat");
+    if !world_dir.is_dir() {
+        // Create directory
```
caelunshun: Consider adding a message here to indicate the world is being created?

server/src/main.rs
```diff
@@ -155,10 +172,68 @@ fn load_config() -> Config {
...
+        thundering: false,
+        thunder_time: 0,
+        time: 0,
+        version: Default::default(),
+        generator_name: config.world.generator.to_string(),
+        generator_options: None,
+    }
+}
+
+fn get_seed(config: &Arc<Config>) -> i64 {
```
caelunshun: Rather than passing a reference to an `Arc`, you can just pass `&Config`—`Arc` dereferences to the wrapped value.

server/src/main.rs
```diff
@@ -155,10 +172,68 @@ fn load_config() -> Config {
...
+    let seed_raw = &config.world.seed;
+    // Empty seed: random
+    // Seed is valid i64: parse
+    // Seed is something else: hash
+    if seed_raw.is_empty() {
+        rand::thread_rng().next_u64() as i64
```
caelunshun: `rand::thread_rng().gen()` should be sufficient here.

caelunshun:
Thanks!

PR #145, opened 2019-10-01
Anvil world persistence

### LINKED ISSUE #52: Anvil world persistence
Currently, worlds can be loaded, but they are not saved.

Resolving this issue will require periodic saving of chunk data, entities, level data, and player data to world saves.

Notes:
* `hematite-nbt` does not support serializing NBT long arrays, which is required for writing chunk data (see PistonDevelopers/hematite_nbt#27). A PR will have to be submitted to them to fix this.

caelunshun:
Resolves #52. 

- [x] Sector allocator for region files
- [x] `RegionFile::save_chunk()`
- [x] Chunk saving support for chunk worker
- [x] Periodic saving of loaded chunks (only those modified), at configurable intervals
- [x] Saving of chunks on unload and shutdown
- [x] Creation of new region files
- [x] Saving of entities
- [x] Saving of player data on disconnect and shutdown
- [x] Saving of level data


caelunshun:
Everything appears to work.

core/src/save/region/mod.rs
```diff
@@ -0,0 +1,834 @@
...
+                                props.into_iter().for_each(|(name, value)| {
+                                    prop_map.insert(name.to_string(), value);
+                                });
+
+                                LevelPaletteEntry {
+                                    name: name.to_string(),
+                                    props: Some(LevelProperties { props: prop_map }),
+                                }
+                            })
+                            .collect(), // TODO: what happens with the global palette?
```
momothereal: I could be wrong, but I don't think there is a "global palette" for local chunk saving.

momothereal:
How is "shutdown" triggered? Is there a hook for sigint/ctrlc that sends a message in the `shutdown` channel?

caelunshun:
> How is "shutdown" triggered? Is there a hook for sigint/ctrlc that sends a message in the `shutdown` channel?

Yes, using the [ctrlc](https://docs.rs/ctrlc/3.1.3/ctrlc/) crate. The shutdown hook is set in `shutdown::init()`.

PR #165, opened 2019-10-16
Implement Bungeecord style IP forwarding

ThijsRay:
This pull request aims to add support for Bungeecord style IP forwarding, as described in https://github.com/caelunshun/feather/issues/155

~~This should not be merged yet, because this branch uses an off-branch mojang-api-rs crate. Pull request https://github.com/caelunshun/mojang-api-rs/pull/6 should be merged first.~~


server/src/io/initialhandler.rs
```diff
@@ -280,14 +368,24 @@ fn handle_login_start(ih: &mut InitialHandler, packet: &LoginStart) -> Result<()
...
-        ih.info = Some(JoinResult {
-            username: ih.username.clone().unwrap(),
-            uuid: Uuid::new_v4(),
-            props: vec![],
-        });
+        let username = packet.username.clone();
+
+        // Check if there is some info about the client available. This can be the case if the
+        // handshake is made by an IP forwarding proxy.
+        if ih.info.is_some() {
```
caelunshun: I believe `if let Some(info) = ih.info.as_mut()` would be more idiomatic here.

ThijsRay: I agree. I have changed it in fe5e9eead5f3009ad708646ffb9c9cb9f37c9038.

caelunshun:
I'm getting a "received BungeeCord data does not match the specification: invalid IP address syntax" error when connecting on localhost. The data sent by BungeeCord is `localhost 127.0.0.1 b26a922dc0604ed5a116e0372a349f1c [{"name":"textures","value":"eyJ0aW1lc3Rh {snip}`. Seems like parsing `localhost` doesn't work correctly, maybe?

ThijsRay:
The type of the host address is `IpAddr`. It seems like `localhost` is not a valid value for this type. I will look into this.

ThijsRay:
> I'm getting a "received BungeeCord data does not match the specification: invalid IP address syntax" error when connecting on localhost. The data sent by BungeeCord is `localhost 127.0.0.1 b26a922dc0604ed5a116e0372a349f1c [{"name":"textures","value":"eyJ0aW1lc3Rh {snip}`. Seems like parsing `localhost` doesn't work correctly, maybe?

I've changed the `IpAddr` type to a regular `String` type to resolve this issue.

caelunshun:
Thanks! Sorry for the delay.

PR #73, opened 2019-08-20
Basic implementation of chat (#57)

momothereal:
Implements handling of the `ChatMessageServerbound` packet, and broadcasting the `ChatMessageClientbound` packet. Largely based on how player animations are implemented.

Closes #57 (unless there is more work in that scope, like chat-components).

PR #173, opened 2020-02-10
Add back basic chat implementation

Wazner:
I noticed the chat implementation was left out from the Legion refactor. I saw an opportunity to get my feet wet with the feather codebase and Legion so I added it back in again.

Changes were based on #73 and #75 

There are no tests, I could not find any examples of other broadcasters with tests.
I can add them if you like, but I would need a hint on where to get started.


caelunshun:
Thanks! This looks good.

Don't worry about tests for now; there isn't a proper way to test systems yet.

PR #174, opened 2020-02-11
Fixed codegen target folder not being ignored

Wazner:
The codegen target folder is not present in the gitignore file.

Some of the other projects are also not present in the gitignore, but I'm not getting a target folder there when running `cargo build`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /blocks/target
 /server/target
 /core/target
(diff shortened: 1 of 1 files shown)


Wazner:
Hmm. These files are actually not generated when you build the workspace. I suppose this PR isn't really needed then, and if we do want to ignore those folders, we should do it for all other projects too.

PR #181, opened 2020-03-19
Heightmaps

Defman:
Heightmap that stores 6 different heights as pr [wiki](https://minecraft.gamepedia.com/Chunk_format)
```rust
pub struct HeightMap {
    motion_blocking: usize,
    motion_blocking_no_leaves: usize,
    ocean_floor: usize,
    ocean_floor_wg: usize,
    world_surface: usize,
    world_surface_wg: usize,
}
```
- [x] Update heightmaps on `set_block_at`
- [x] Save heightmaps
- [ ] Load heightmaps
- [ ] World generated heightmap, always stays at 0

I have added `chunk.recalculate_heightmaps()` to chunk generation, not sure if it is correct approch.

### DIFF
--- a/core/src/chunk.rs
+++ b/core/src/chunk.rs
@@ -59,6 +59,67 @@ pub struct Chunk {
     /// Whether this chunk has been modified since the most recent
     /// call to `check_modified`().
     modified: bool,
+
+    heightmaps: [HeightMap; CHUNK_WIDTH * CHUNK_WIDTH],
+}
+
+#[derive(Clone, Copy, Default)]
+pub struct HeightMap {
+    motion_blocking: usize,
+    motion_blocking_no_leaves: usize,
+    ocean_floor: usize,
+    ocean_floor_wg: usize,
+    world_surface: usize,
+    world_surface_wg: usize,
+}
+
+impl HeightMap {
+
+    /// The highest block that is solid or contains a fluid.
+    pub fn motion_blocking(&self) -> usize {
+        self.motion_blocking
+    }
+
+    pub fn set_motion_blocking(&mut self, motion_blocking: usize) {
+        self.motion_blocking = motion_blocking;
+    }
+
+    /// The highest block that is solid or contains a fluid and is not leaves.
+    pub fn motion_blocking_no_leaves(&self) -> usize {
+        self.motion_blocking_no_leaves
+    }
+
+    pub fn set_motion_blocking_no_leaves(&mut self, motion_blocking_no_leaves: usize) {
+        self.motion_blocking_no_leaves = motion_blocking_no_leaves;
+    }
+
+    /// The highest block that is solid.
+    pub fn ocean_floor(&self) -> usize {
+        self.ocean_floor
+    }
+
+    pub fn set_ocean_floor(&mut self, ocean_floor: usize) {
+        self.ocean_floor = ocean_floor;
+    }
+
+    /// The highest block that is solid for world generation.
+    pub fn ocean_floor_wg(&self) -> usize {
+        self.ocean_floor_wg
+    }
+
+    /// The highest block that is not air.
+    pub fn world_surface(&self) -> usize {
+        self.world_surface
+    }
+
+    pub fn set_world_surface(&mut self, world_surface: usize) {
+        self.world_surface = world_surface;
+    }
+
+    /// The highest block is not air for world generation.
+    pub fn world_surface_wg(&self) -> usize {
+        self.world_surface_wg
+    }
 }
 
 impl Default for Chunk {
@@ -77,6 +138,7 @@ impl Default for Chunk {
             modified: true,
             sections,
             biomes: [Biome::Plains; SECTION_WIDTH * SECTION_WIDTH],
+            heightmaps: [HeightMap::default(); CHUNK_WIDTH * CHUNK_WIDTH],
         }
     }
 }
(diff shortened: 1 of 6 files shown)


core/src/chunk.rs
```diff
@@ -151,6 +213,58 @@ impl Chunk {
...
+
+    /// Recalculate the heightmap for the chunk
+    pub fn recalculate_heightmap(&mut self) {
+        // This function can be optimized, instead of 
+        // fetching heightmap every time, and sections
+        for x in 0..CHUNK_WIDTH {
+            for z in 0..CHUNK_WIDTH {
+                for y in (0..CHUNK_HEIGHT).rev() {
+                    let block = self.block_at(x, y, z);
+                    self.update_heightmap(x, y, z, block);
```
caelunshun: Could we have `update_heightmap` return a bool that indicates whether it encountered a block? This way, we could skip checking the rest of the y values because we found the highest block

core/src/save/region/mod.rs
```diff
@@ -344,6 +348,14 @@ fn read_section_into_chunk(section: &LevelSection, chunk: &mut Chunk) -> Result<
 }
 
 fn chunk_to_chunk_root(chunk: &Chunk, entities: Vec<EntityData>) -> ChunkRoot {
+    let heightmaps: Vec<i64> = chunk.heightmaps().iter().map(|map| {
+        (map.motion_blocking() << (9 * 0)
```
caelunshun: Could 9 be stored in a constant here?

caelunshun:
Thanks, looks good! Just a couple comments.

core/src/chunk.rs
```diff
@@ -59,6 +59,67 @@ pub struct Chunk {
...
+
+#[derive(Clone, Copy, Default)]
+pub struct HeightMap {
+    motion_blocking: usize,
+    motion_blocking_no_leaves: usize,
+    ocean_floor: usize,
+    ocean_floor_wg: usize,
+    world_surface: usize,
+    world_surface_wg: usize,
+}
```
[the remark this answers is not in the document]
Defman: Storing as arrays of `u8` would lower the strain on the CPU cache.

Started with `u8` but turned it into `usize` due to all y coordinates being `usize` in every other function.

core/src/chunk.rs
Defman: Sure, that would be nice; however, still need to figure out how to recalculate the world generated heights, ie. `world_surface_wg` and `ocean_floor_wg`.

core/src/save/region/mod.rs
Defman: Definitely, however, I'm not sure if I should move the serialization and deserialization to somewhere else.

core/src/chunk.rs
```diff
@@ -59,6 +60,77 @@ pub struct Chunk {
...
+    world_surface_wg: u8,
+}
+
+impl HeightMap {
+    /// The highest block that is solid or contains a fluid.
+    pub fn motion_blocking(self) -> u8 {
+        self.motion_blocking
+    }
+
+    pub fn set_motion_blocking(mut self, motion_blocking: u8) {
```
caelunshun: Not sure why you're taking in `mut self`, as this passes ownership of the `HeightMap` to the function and any changes get discarded.

core/src/chunk.rs
```diff
@@ -59,6 +60,77 @@ pub struct Chunk {
...
+pub struct HeightMap {
+    motion_blocking: u8,
+    motion_blocking_no_leaves: u8,
+    ocean_floor: u8,
+    ocean_floor_wg: u8,
+    world_surface: u8,
+    world_surface_wg: u8,
+}
+
+impl HeightMap {
```
MarkMcCaskey: Because `HeightMap` is Copy and just plain data, I think the idiomatic thing to do in rust is to make the fields public and remove all getters and setters.  Mutability is already controlled in Rust and if the getters and setters aren't upholding invariants then they don't really serve any purpose at all in Rust.

core/src/save/region/mod.rs
```diff
@@ -344,6 +351,18 @@ fn read_section_into_chunk(section: &LevelSection, chunk: &mut Chunk) -> Result<
 }
 
 fn chunk_to_chunk_root(chunk: &Chunk, entities: Vec<EntityData>) -> ChunkRoot {
+    let heightmaps: Vec<i64> = chunk
+        .heightmaps()
+        .iter()
+        .map(|map| {
+            (map.motion_blocking() as i64)
```
MarkMcCaskey: ⛳️ 

Perhaps there should be `From`/`Into` Implementations for HeightMap -> i64 and i64 -> HeightMap

core/src/chunk.rs
MarkMcCaskey: The doc comments on these functions are great though, those should be moved to the struct fields! (if the no-op getters/setters are removed)

Defman: You should not be able to modify world geneated heights.

core/src/save/region/mod.rs
Defman: Same thought, however, would I not be moving serlization details into heightmap?

Defman: However maybe we want to be able to modify the generated height at some point, and it should be up to the user if she/he has the need?

core/src/chunk.rs
Defman: However maybe we want to be able to modify the generated height at some point, and it should be up to the user if she/he has the need?

Defman: After some thought, I realized that you should not be able to obtain a mutable reference of a Heightmap and therefore all fields should be public.

PR #185, opened 2020-03-20
Add support for shooting arrows

Redrield:
Physics for the entity is still a bit weird once it's spawned, however, this code successfully handles spawning the entity when the bow is released.

### DIFF
--- a/server/src/entity/arrow.rs
+++ b/server/src/entity/arrow.rs
@@ -1,0 +1,62 @@
+use crate::entity;
+use crate::entity::{ComponentSerializer, EntityId, SpawnPacketCreator, Velocity};
+use crate::game::Game;
+use crate::physics::PhysicsBuilder;
+use crate::util::{degrees_to_stops, protocol_velocity};
+use feather_core::entity::{ArrowEntityData, BaseEntityData, EntityData};
+use feather_core::network::packet::implementation::SpawnObject;
+use feather_core::{Packet, Position, Vec3d};
+use fecs::{EntityRef, World};
+use uuid::Uuid;
+
+pub fn spawn_arrow(game: &mut Game, world: &mut World, velocity: glm::DVec3, position: Position) {
+    let entity = entity::base(position)
+        .with(Velocity(velocity))
+        .with(SpawnPacketCreator(&create_spawn_packet))
+        .with(ComponentSerializer(&serialize))
+        .with(
+            PhysicsBuilder::new()
+                .bbox(0.5, 0.5, 0.5)
+                .gravity(-0.05)
+                .slip_multiplier(0.0)
+                .drag(0.99)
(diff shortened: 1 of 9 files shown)


server/src/packet_handlers/digging.rs
```diff
@@ -210,41 +193,94 @@ fn handle_shoot_bow(
...
+        arrow_velocity.norm()
+    );
+
+    drop(timed_use);
+
+    world.remove::<ItemTimedUse>(player).unwrap();
+
+    debug!("Preparing to spawn entity");
+
+    let entity = entity::base(init_position)
```
caelunshun: This code as well as the `create_spawn_packet` function and the `serialize` function should be probably moved into `entity/arrow.rs` to keep consistent with the item entity implementation.

server/src/packet_handlers/digging.rs
```diff
@@ -210,41 +193,94 @@ fn handle_shoot_bow(
...
+
+    let direction = init_position.direction();
+
+    let arrow_velocity = compute_projectile_velocity(
+        glm::vec3(direction.x, direction.y, direction.z),
+        charge_force as f64,
+        0.0,
+        &mut *game.rng(),
+    );
+    debug!(
```
caelunshun: These should probably be trace messages

server/src/packet_handlers/use_item.rs
```diff
@@ -0,0 +1,55 @@
...
+        world
+            .add(
+                player,
+                ItemTimedUse {
+                    tick_start: game.tick_count,
+                },
+            )
+            .unwrap();
+        let player_name = world.get::<Name>(player);
+        info!("Added ItemTimedUse to player {}.", player_name.0);
```
caelunshun: This should be a trace message as well.

caelunshun:
Thanks, excellent work on this! A few small fixes and this should be good to go.

server/src/entity/arrow.rs
```diff
@@ -0,0 +1,62 @@
...
+use crate::game::Game;
+use crate::physics::PhysicsBuilder;
+use crate::util::{degrees_to_stops, protocol_velocity};
+use feather_core::entity::{ArrowEntityData, BaseEntityData, EntityData};
+use feather_core::network::packet::implementation::SpawnObject;
+use feather_core::{Packet, Position, Vec3d};
+use fecs::{EntityRef, World};
+use uuid::Uuid;
+
+pub fn spawn_arrow(game: &mut Game, world: &mut World, velocity: glm::DVec3, position: Position) {
```
caelunshun: To be consistent with the item implementation, this should be called `create` and return an `EntityBuilder` for adding further components.

caelunshun:
Thanks!

PR #186, opened 2020-03-21
Text API: WIP

Defman:
Feedback: Am I going the right direction?

### DIFF
--- a/core/src/text.rs
+++ b/core/src/text.rs
@@ -44,7 +44,6 @@ impl From<Style> for Text {
     }
 }
 
-// TODO: use instead of string as keybind
 #[derive(Debug, PartialEq)]
 pub enum KeyBind {
     Attack,
@@ -283,7 +282,6 @@ pub enum Hover {
     ShowEntity(String),
 }
 
-// TODO: Transform into string, fn(self, translator: Translator) -> String
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum TextValue {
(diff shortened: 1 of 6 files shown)


core/src/text.rs
```diff
@@ -0,0 +1,461 @@
...
+    #[serde(rename = "show_entity")] 
+    // TODO: Entity struct
+    ShowEntity(String),
+}
+
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum TextValue {
+    Text {
+        text: String,
```
caelunshun: Could we use `Cow<'static, str>` to allow either static strings without allocation or heap-allocated `String`s? Functions which take an `&str` could then be switched to take `impl Into<Cow<'static, str>>`.

caelunshun:
This looks great, thanks for the dedicated work! I think you're on the right path here—I can't see anything which could become a problem in the future.

Defman:
Current only `TextComponent` support the builder patter, ie. TextComponent.red() should `Text` also support this? Current `Text` supports `Text::from("hello world") * Color::Red`.

Defman:
The translate enum should also contain all translation keys for 1.13.2.

caelunshun:
Just did a thorough review—phenomenal work. A few suggestions/questions:
* Could we add some docs which explain the differences between `Text`, `TextComponent`, `TextValue`, and `TextRoot`? I feel this could be somewhat confusing for new developers.
* Also, there seem to be two ways to build `Text` objects: using the operator overloading style API and the `TextComponentBuilder` API. What are the key differences between these two designs? If they both do the same thing, could we decide on just one of them, for simplicity?

Defman:
Operator overloading is supposed to be a wrapper around `TextComponentBuilder`, such that
```rust
impl<T> std::ops::Mul<Color> for T
where
    T: TextComponentBuilder,
```

PR #190, opened 2020-03-26
TextAPI: Refactor

Defman:
- Split the massive text.rs into multiple logical files and minor improvements.
- I have chosen to remove TextRoot since it should be handle at the consumer end.
- Started on simplification/normalization of text such that it is possible to compare if two texts are equivalent.
- Started on language translation API, such that it is possible to apply translation to a translatable text. 
- Changes to how text is sent such that it enforces what was previously delegated to `RootText`.
- Added more tests.

Note: Json-NBT wiki.vg seems to suggest that it is not a regular JSON format, so the serialization might have to change at places.
```json
"value": {
  "text": "{id:35,Damage:5,Count:2,tag:{display:{Name:Testing}}}"
}
```
instead of
```json
"value": {
  "text": {
    "id": 35,
    "Damage": 5,
    "Count": 2,
    "tag": {
      "display": {
        "Name": "Testing"
      }
    }
  }
}
```


core/src/text/language.rs
```diff
@@ -0,0 +1,28 @@
+use crate::text::{Text, Translate};
+
+pub trait Language {
+    fn translate_with<I>(&self, translate: Translate, with: I) -> Text
+    where
+        I: IntoIterator,
+        I::Item: Into<Text>;
+}
+
+impl Language for std::collections::HashMap<Translate, String> {
```
caelunshun: Could this be made generic over the hasher parameter `S`? https://rust-lang.github.io/rust-clippy/master/#implicit_hasher

caelunshun:
@Defman is this ready for merging?

Defman:
For the most part, yes. However some none essential parts still need work.
- Text simplification such it is possible to compare the equality of Text.
- Languages, such that it is possible to turn translatable text into actual text.
- The Translate enum should be generated from feather-data/lang/EN_US

We are most likely also going to discover, that I misunderstood some parts of Minecraft raw text. I would also like the API to adhere to some common pattern throughout feather. In other words, does the API look useful and decent?

Defman:
@caelunshun I think we should revisit this PR, it contains quite a few imrpovements but it still have some none critical missing, like simplification and translation.

PR #191, opened 2020-03-27
Minecraft data: Minecraft data generater

Defman:
Downloads the Minecraft server jar and uses the built-in generate to obtain useful data. it also extracts useful data from the jar.


Defman:
I'm not sure about the location of the downloaded jar and generated data.

data/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+#[cfg(test)]
```
caelunshun: Should probably remove this stub?

.gitignore
```diff
@@ -1,9 +1,10 @@
 /target
-/blocks/target
-/server/target
-/core/target
+**/target
+**/target
```
caelunshun: I think we only need one `**/target` entry.

Defman:
I have made it such feather-data generates a module for the embedded data as such.
```rust
feather_data::minecraft::lang::EN_US;
feather_data::minecraft::BLOCKS;
feather_data::minecraft::data::loot_tables::entities::BAT;
``

caelunshun:
Thanks, nice work on this.

PR #188, opened 2020-03-23
Fix CI

caelunshun:
* Stop running builds on Windows and macOS as they fail with obscure linker errors.

Fixes #171

PR #192, opened 2020-03-29
Disabled jemalloc for msvc toolchain

terafo:
Feather build fails on msvc toolchain because of jemalloc incompatibilities.


Defman:
#188 Already fixes this

Defman:
I would suggest that you request changes to the CI PR(#188) and submit a PR that fixes the Unicode decoding of strings.

PR #195, opened 2020-04-04
Fixed unicode encoding issues in chat

terafo:
Previously any unicode symbol was split to 8 bir chars. Now it's fixed.

### DIFF
--- a/core/src/network/mctypes.rs
+++ b/core/src/network/mctypes.rs
@@ -159,14 +159,14 @@ impl<B: Buf + Read> McTypeRead for B {
                 return Err(TryGetError::ValueTooLarge);
             }
             let mut result = String::with_capacity(len as usize);
-            for _ in 0..len {
-                let res = self.try_get_i8()? as u8;
-                result.push(res as char);
-            }
-
+            let _ = self
+                .take(len as u64)
+                .read_to_string(&mut result)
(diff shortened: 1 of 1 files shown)


Defman:
In reviewing the code I realized that take only guarantee that at most `n` bytes will be read, that the reader actually contains enough bytes. Therefore we must verify the number of reading bytes is equal to the expected length. The var_int for the length is at most `3` bytes as well.

```rust
/// Reads a string from the object.
fn try_get_string(&mut self) -> Result<String, TryGetError> {
    // VarInt is at most 3 bytes.
    let len = self.take(3).get_mut().try_get_var_int()?;
    // Check that the client isn't trying
    // to make the server allocate ridiculous
    // amounts of memory
    if len > 32767 {
        return Err(TryGetError::ValueTooLarge);
    }
    let mut result = String::with_capacity(len as usize);
    let read_len = self
        .take(len as u64)
        .read_to_string(&mut result)
        .map_err(|_| TryGetError::InvalidValue(-1))?;
    // Verify the number of bytes actually read.
    if read_len == len as usize {
        Ok(result)
    } else {
        Err(TryGetError::NotEnoughBytes)
    }
}
```

caelunshun:
Thanks! As there were some merge conflicts here I applied this in 8614692f8e0a24979853e29d41567b5a82249831, along with Defman's suggestion.

PR #198, opened 2020-04-21
Destroy non-solid blocks when the block underneath is broken (#135)

Redrield:


server/src/packet_handlers/digging.rs
```diff
@@ -13,8 +13,12 @@ use crate::player::{ItemTimedUse, PLAYER_EYE_HEIGHT};
 use feather_blocks::BlockId;
 use feather_core::inventory::{SlotIndex, SLOT_HOTBAR_OFFSET, SLOT_OFFHAND};
 use feather_core::network::packet::implementation::{PlayerDigging, PlayerDiggingStatus};
+<<<<<<< HEAD
```
caelunshun, line 16: Looks like there's some unresolved merge conflict here

PR #212, opened 2020-05-04
Text formatting markup language

Redrield:
Add support for parsing a simple markdown language into TextComponents

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,8 @@ members = [
     "core/items",
     "core/inventory",
     "core/network",
-    "core/util",
     "core/text",
+    "core/util",
 
     "server",
     "server/template",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ members = [
     "core/inventory",
     "core/network",
     "core/util",
+    "core/text",
 
     "server",
     "server/template",
--- a/core/text/Cargo.toml
+++ b/core/text/Cargo.toml
@@ -12,3 +12,4 @@ serde_json = "1.0"
 serde_with = "1.4"
 hematite-nbt = "0.4"
 uuid = { version = "0.8", features = ["serde"] }
+nom = "5.1.1"
--- a/core/text/Cargo.toml
+++ b/core/text/Cargo.toml
@@ -12,4 +12,5 @@ serde_json = "1.0"
 serde_with = "1.4"
 hematite-nbt = "0.4"
 uuid = { version = "0.8", features = ["serde"] }
-nom = "5.1.1"
+nom = "5.1"
+thiserror = "1.0"
--- a/core/text/Cargo.toml
+++ b/core/text/Cargo.toml
@@ -13,4 +13,5 @@ serde_with = "1.4"
 hematite-nbt = "0.4"
 uuid = { version = "0.8", features = ["serde"] }
 nom = "5.1"
+nom_locate = "2.0"
 thiserror = "1.0"
--- a/core/text/src/markdown/lexer.rs
+++ b/core/text/src/markdown/lexer.rs
@@ -1,13 +1,13 @@
-use nom::{IResult, InputLength, InputTake, InputIter, Slice};
-use std::ops::{Range, RangeFrom, RangeTo, RangeFull};
-use std::slice::Iter;
-use std::iter::Enumerate;
-use nom::multi::*;
 use nom::branch::*;
-use nom::combinator::*;
 use nom::bytes::complete::*;
 use nom::character::complete::*;
+use nom::combinator::*;
+use nom::multi::*;
 use nom::sequence::*;
+use nom::{IResult, InputIter, InputLength, InputTake, Slice};
+use std::iter::Enumerate;
+use std::ops::{Range, RangeFrom, RangeFull, RangeTo};
+use std::slice::Iter;
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum LexToken {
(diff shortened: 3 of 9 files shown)


core/text/src/markdown/lexer.rs
```diff
@@ -0,0 +1,215 @@
...
+use nom::bytes::complete::*;
+use nom::character::complete::*;
+use nom::sequence::*;
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum LexToken {
+    ControlWordStarter,
+    LBrace,
+    RBrace,
+    Text(String),
```
Defman, line 17: This is not part of the lexing
Tokens are the following
- WhiteSpace
- Word
- LBrace
- RBrace
- ControlWordStarter

Text is an internal literal

core/text/src/markdown/lexer.rs
```diff
@@ -0,0 +1,215 @@
...
+    ColorCode(String),
```
Defman, line 18: Color codes are just a word Word("#00FFAA").

Cargo.toml
```diff
@@ -15,6 +15,7 @@ members = [
     "core/inventory",
     "core/network",
     "core/util",
+    "core/text",
```
caelunshun, line 18: This should follow alphabetical order (put it directly above core/util)

core/text/Cargo.toml
```diff
@@ -12,3 +12,4 @@ serde_json = "1.0"
 serde_with = "1.4"
 hematite-nbt = "0.4"
 uuid = { version = "0.8", features = ["serde"] }
+nom = "5.1.1"
```
caelunshun, line 15: Our convention is to only specify the major and minor version, i.e. `nom = "5.1"`

core/text/src/markdown/lexer.rs
```diff
@@ -0,0 +1,218 @@
...
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub struct Tokens<'a> {
+    pub tok: &'a [LexToken],
+    start: usize,
+    end: usize,
+}
+
+impl<'a> Tokens<'a> {
+    #[allow(clippy::ptr_arg)]
+    pub fn new(vec: &'a Vec<LexToken>) -> Tokens<'a> {
```
caelunshun, line 30: This function should be able to take a slice instead of a `Vec`, if I'm not mistaken?

core/text/src/markdown/lexer.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+    }
+}
+
+pub fn lex_control_word(input: &str) -> IResult<&str, LexToken> {
+    map(tag("@"), |_| LexToken::ControlWordStarter)(input)
+}
+
+pub fn lex_spaces(input: &str) -> IResult<&str, LexToken> {
+    map(space1, |s: &str| LexToken::Space(s.to_string()))(input)
```
caelunshun, line 143: We could reduce allocations by storing string slices rather than owned `String`s, though it unlikely matters in practice. It seems fairly trivial to do in this case, however.

core/text/src/markdown/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+}
+
+pub fn has_lbrace(input: Tokens) -> isize {
+    for (i, tok) in input.tok.iter().enumerate() {
+        if let LexToken::LBrace = tok {
+            return i as isize;
+        }
+    }
+
+    -1isize
```
caelunshun, line 68: It's more idiomatic here to return an `Option::None` if no brace is found.

core/text/src/markdown/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+}
+
+pub fn parse_control_word(i: Tokens) -> IResult<Tokens, Token> {
+    let (i, _) = token(LexToken::ControlWordStarter)(i)?;
+    let (i, next) = take(1usize)(i)?;
+
+    match &next.tok[0] {
+        LexToken::Word(s) => match s.as_str() {
+            "red" => {
```
caelunshun, line 77: I'm sure you're already planning to fix this, but to eliminate duplicate code there should be some enum to determine the formatting specifier in the input to replace the duplicated match arms.

core/text/src/markdown/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+                }
+            }
+            "color" => {
+                let (_i, next) = take(1usize)(i)?;
+                match &next.tok[0] {
+                    LexToken::Word(_cc) => todo!("ColorCode into Color"),
+                    _ => panic!("Error branch"),
+                }
+            }
+            _ => panic!("Error branch"),
```
caelunshun, line 132: shouldn't this be handled gracefully?

core/text/src/markdown/parser.rs
```diff
@@ -0,0 +1,280 @@
...
+                        style: Style::Bold,
+                        rest: vec![Token::Text(
+                            "Some red bold text and more red bold text".to_string()
+                        )]
+                    }),
+                ]
+            }),]
+        );
+    }
+}
```
caelunshun, line 280: This parser would be a good use case for fuzz testing, to find any panics lying around.

caelunshun:
Nice work, just a few minor nitpicks!

core/text/src/markdown/lexer.rs
Redrield: Normally yes, however it's an internal function and the only place that it is used is internally where the only sequence of LexTokens that are encountered are `Vec`s from `lex_input`.

caelunshun:
Thanks!

PR #213, opened 2020-05-05
Initial plugin support

### LINKED ISSUE #106: Implement a plugin system
This is a key component in making Feather usable.

Requirements:
* Implement a system which dynamically loads plugins at runtime. Plugins should be able to handle events and register `System`s.
* Plugins should be able to access and modify components.
* The boundary between plugins and server should be efficient, ideally zero-cost.
* Plugins should be written in Rust.

The proposal is to compile plugins to WebAssembly and run them using [wasmer](https://github.com/wasmerio/wasmer). At server startup, plugins will be loaded and an `init` function will be called. Here, plugins will be able to register systems and event handlers with the host.

Any systems registered by a plugin will be added to the `DispatcherBuilder` at init time. This means that plugin systems will run in the same way normal ones do. 

Plugins will be required to specify the system data they require for each system, much like systems on the host, but components and resources will be represented in a more abstract way. For example, they will not have to manually write events; rather, this will be handled transparently on the host.

Any requests and updates of data by a plugin will require a host call. The host exposes functions to query and update data. On the host side, any `SystemData` required by the plugin will be stored in `wasmer::Ctx`, and inside a host function, the function will take the `SystemData` from the `Ctx` and use it as necessary. The exact details in how this will be implemented will be investigated later, but we know it is possible.

The plugin API itself will look much like the Bukkit API at first glance, with the main difference being the requirement for specifying what data an event handler requires. In addition, plugins will be allowed to have a state struct which is also passed to any event handler or system. Using this, it can access any necessary global state. The state struct will be required to be `Send + Sync`, since in the future, plugins may be invoked in parallel (although currently, this is not possible since `wasmer::Instance` does not implement `Sync`).

Several types from `feather-core` will have to be moved to the plugin API, such as `Position`. Also, `feather-blocks` and `feather-items` will be exposed to plugins.

The current plan is to first implement a proof of concept using Specs with Wasmer before proceeding to write the actual plugin system. Work on this should begin within the next few months.

caelunshun:
This PR will implement an initial design for `fapi`, our plugin API, and the associated API bindings and plugin loader in `feather-server`.

The current design is based on dynamic loading of shared libraries using a C-compatible FFI interface. The decision was made to hold off on the use of WebAssembly as the WASM ecosystem lacks the maturity we desire. For example, networking is still unimplemented in WASI (and has been for well over a year since the issue was filed...). Networking is a critical feature for plugins needing database interaction.

The current API can be ported to WASM with a bit of effort, should WASI implement the features we need.

Resolves #106, at least initially.

### Checklist

- [x] host vtable abstraction + macro
- [x] `plugin_init` proc macro to mark a plugin's initialization function. (Question: this naming isn't ideal; ideally it would be called `plugin`, but that's taken by the Rust compiler. Other ideas are welcome)
- [x] Plugin logging
- [ ] Test framework to test the API and server implementation
* [ ] Registration of systems
* [ ] Registration of event handlers
* [ ] Interaction with entities and the world
* [ ] Custom components
* [ ] Plugin state
* [ ] Interaction between plugins
* [ ] Documentation
* [ ] Release `fapi` to crates.io

### DIFF
--- a/api/common/src/system.rs
+++ b/api/common/src/system.rs
@@ -1,0 +1,23 @@
+//! Defines types for registering systems and event handlers.
+
+use crate::event::{EventId, OpaqueEvent};
+use crate::states::GameState;
+
+/// Specifies a description of a system.
+#[repr(C)]
+pub struct SystemSpec {
+    /// A pointer to the function to run when the system is invoked.
+    pub f: fn(*mut GameState),
+}
+
+/// Specifies a description of an event handler.
+#[repr(C)]
+pub struct HandlerSpec {
+    /// The function to invoke when the handler runs.
+    ///
+    /// The function may assume the `OpaqueEvent` pointer
+    /// points to the event type associated with `event`.
+    pub f: fn(*mut GameState, *mut OpaqueEvent),
+    /// The ID of the event type which is to be handled by `f`.
+    pub event: EventId,
+}
--- a/api/common/src/util.rs
+++ b/api/common/src/util.rs
@@ -1,0 +1,78 @@
+use bstr::BStr;
+use std::alloc::Layout;
+use std::marker::PhantomData;
+use std::slice;
+
+/// A string reference which may be passed over FFI boundaries.
+///
+/// Does not necessarily point to valid UTF8.
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct FStr<'a> {
+    ptr: *const u8,
+    len: usize,
+    _lifetime: PhantomData<&'a [u8]>,
+}
+
+impl<'a> From<&'a BStr> for FStr<'a> {
+    fn from(s: &'a BStr) -> Self {
+        let slice: &[u8] = s.as_ref();
+        Self::from(slice)
+    }
+}
+
+impl<'a> From<&'a [u8]> for FStr<'a> {
+    fn from(slice: &'a [u8]) -> Self {
+        Self {
+            ptr: slice.as_ptr(),
+            len: slice.len(),
+            _lifetime: PhantomData,
+        }
+    }
+}
+
+impl<'a> From<&'a str> for FStr<'a> {
+    fn from(string: &'a str) -> Self {
+        Self {
+            ptr: string.as_ptr(),
+            len: string.len(),
+            _lifetime: PhantomData,
+        }
+    }
+}
+
+impl<'a> From<FStr<'a>> for &'a BStr {
+    fn from(string: FStr<'a>) -> Self {
+        let slice = <&[u8]>::from(string);
+        slice.into()
+    }
+}
+
+impl<'a> From<FStr<'a>> for &'a [u8] {
+    fn from(string: FStr<'a>) -> Self {
+        unsafe { slice::from_raw_parts(string.ptr, string.len) }
+    }
+}
+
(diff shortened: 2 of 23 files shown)


api/src/vtable.rs
```diff
@@ -0,0 +1,29 @@
...
+use fapi_common::vtable::HostVTable;
+
+static mut VTABLE: Option<HostVTable> = None;
+
+/// Intializes the vtable from the given string.
+///
+/// # Safety
+/// This function may not be called from multiple threads
+/// at once.
+pub fn init_vtable(s: FStr) -> Result<(), serde_json::Error> {
```
Defman, line 11: Could we enforce this by having a static lock?

api/common/src/system.rs
```diff
@@ -0,0 +1,23 @@
+//! Defines types for registering systems and event handlers.
+
+use crate::event::{EventId, OpaqueEvent};
+use crate::states::GameState;
+
+/// Specifies a description of a system.
+#[repr(C)]
+pub struct SystemSpec {
+    /// A pointer to the function to run when the system is invoked.
+    pub f: fn(*mut GameState),
```
Defman, line 10: Can we change the name to something more descriptive than `f`, like `run`, `invoke`, or `step`?

api/common/src/system.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+/// Specifies a description of an event handler.
+#[repr(C)]
+pub struct HandlerSpec {
+    /// The function to invoke when the handler runs.
+    ///
+    /// The function may assume the `OpaqueEvent` pointer
+    /// points to the event type associated with `event`.
+    pub f: fn(*mut GameState, *mut OpaqueEvent),
```
Defman, line 20: Better field name like, `handle`, `emit`, `run`?

api/common/src/util.rs
```diff
@@ -0,0 +1,78 @@
...
+
+/// A string reference which may be passed over FFI boundaries.
+///
+/// Does not necessarily point to valid UTF8.
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct FStr<'a> {
+    ptr: *const u8,
+    len: usize,
+    _lifetime: PhantomData<&'a [u8]>,
```
Defman, line 14: is `[u8]` prefered over the `()` unit type?

api/src/vtable.rs
caelunshun: Cost will likely be significant, since this is hot code. This is an internal API, so there's not too much need for enforcing safety, as long as we're careful. For now, though, I can switch this to a safe implementation.

api/common/src/util.rs
caelunshun: The type of PhantomData is supposed to indicate what the phantom lifetime is used for—in this case, the lifetime is used for `ptr`, which is a pointer to a slice of `u8`.

api/src/vtable.rs
Redrield: If the function has safety implications for calling, wouldn't it be better to make it an `unsafe fn` rather than wrap that up with a note in the docs?

caelunshun:
Superseded by #307 and #349.

PR #216, opened 2020-05-08
Added propper error message for release mode on failed authentication

Defman:
We should most likely change how disconnects are handle in generally, such that we can support raw json text.


Defman:
This also raises the question, should we just use `dbg` or `trace` for errors or should we emit the error through the disconnect message?

caelunshun:
Thanks!

PR #221, opened 2020-05-12
Implement a new inventory API

### LINKED ISSUE #79: Correctly handle inventory indexing

caelunshun:
The old inventory design had a key problem: It could not correctly handle slot indexing in different contexts. As per [this page](https://wiki.vg/Inventory#Player_Inventory), indices in the protocol take different forms depending on the current window. The old API used raw indices based on those used for the main inventory context, but this cannot be extended to work with other windows, such as crafting tables, chests, etc.

The new design hopes to solve the above.

Resolves #79.


Defman:
How are we going to handle serialization?

Defman:
Also, which kinds of operations do we expect users to have a need for?
The ability to subtract and add a number of items based on attributes? Ie, type, tags, durability, and etc?

Would these operations be carried out on the entire inventory or a subset, ie hotbar and armor?

caelunshun:
@Defman for querying, here's some use cases I can think of:

* Querying for an empty slot, or a slot with a certain item, with configurable order. For example, the `collect_item` function currently does this.
* Querying for some quantity of an item of a set of types—the arrow shooting system currently does this to remove shot arrows from the inventory.

PR #223, opened 2020-05-14
Send game state before PlayerJoinEvent

### LINKED ISSUE #217: Initial game state is sent after Join Game packet
The initial game state—time, weather, inventory—is currently sent after the Join Game packet. This leads to a brief period upon joining the server during which the time is dawn and the weather is clear, regardless of the actual values.

To solve this, there should be a new event—e.g. `PlayerPreJoinEvent`—triggered before the join game packet is sent. The event handlers which send initial game state can then be switched to handle the new event.

nikhiljha:
closes #217 

I'm trying to do some small fixes to start.


caelunshun:
Thanks, sorry about the delay!

PR #227, opened 2020-05-16
Fix crash caused by accessing block above 255

Defman:


caelunshun:
Thanks! Gotta love those off-by-one errors...

PR #230, opened 2020-05-18
Attempt to improve CI build times

caelunshun:


caelunshun:
Am proud to announce that CI now spends more time on caching the build than actually building the project.

Issue #229, opened 2020-05-18
Tracking issue: complete command support


caelunshun:
#209 added initial support for commands using [`lieutenant`](https://github.com/feather-rs/lieutenant) and implemented `/tp` and `/gamemode`. What's left to do:

* Implement [all commands in Vanilla](https://minecraft.gamepedia.com/Commands#List_and_summary_of_commands) (only those present in Java Edition)
* Ensure that messages are consistent with Vanilla (this is certainly not the case right now)
* Implement [target selector arguments](https://minecraft.gamepedia.com/Commands#Target_selector_arguments)
* [Send the command node graph to players upon joining](https://wiki.vg/Protocol#Declare_Commands)

For newcomers to the project, implementing a command is a good way to start. The basic process is:
* Locate the Minecraft Wiki page for the command and read through it, so you understand the command well (skip any Bedrock-specific sections).
* Go into `server/commands/src/impls.rs` and implement the command.
  * The `tp` implementation is a good example of how a command could be implemented.
  * If you need a new argument type (for example, a parsed gamemode or a UUID), implement it in `server/commands/src/arguments.rs`. `EntitySelector` and `ParsedGamemode` are useful examples to see how this is done.

If you need any help, you are welcome to ask as many questions as you need on our Discord server :)

amberkowalski:
I'm currently working on the [tell](https://minecraft.gamepedia.com/Commands/tell) command.

onbjerg:
Java Edition vanilla commands list:

- [ ] [`/attribute`](https://minecraft.gamepedia.com/Commands/attribute)
- [ ] [`/locatebiome`](https://minecraft.gamepedia.com/Commands/locatebiome)
- [ ] [`/advancement`](https://minecraft.gamepedia.com/Commands/advancement)
- [ ] [`/ban`](https://minecraft.gamepedia.com/Commands/ban)
- [ ] [`/ban-ip`](https://minecraft.gamepedia.com/Commands/ban-ip)
- [ ] [`/banlist`](https://minecraft.gamepedia.com/Commands/banlist)
- [ ] [`/bossbar`](https://minecraft.gamepedia.com/Commands/bossbar)
- [ ] [`/clear`](https://minecraft.gamepedia.com/Commands/clear)
- [ ] [`/clone`](https://minecraft.gamepedia.com/Commands/clone)
- [ ] [`/data`](https://minecraft.gamepedia.com/Commands/data)
[...]

PR #232, opened 2020-05-19
Add /say command

amberkowalski:
Adds /say command for #229.

### DIFF
--- a/server/commands/src/impls.rs
+++ b/server/commands/src/impls.rs
@@ -191,19 +191,22 @@ pub fn whisper(
 
 #[command(usage = "say <message>")]
 pub fn say(ctx: &mut CommandCtx, message: TextArgument) -> anyhow::Result<()> {
-    let sender_name = if let Some(sender_name) = ctx.world.try_get::<Name>(ctx.sender) {
-        sender_name.0.clone()
+    let name = ctx.world.try_get::<Name>(ctx.sender);
+
+    let sender_name = if let Some(name) = &name {
+        &name.0.as_str()
     } else {
-        // Use a default value if the executor has no Name component
-        String::from("Server")
+        "Server"
     };
 
-    let command_output = Text::from(format!("[{}] {}", sender_name, message.0)).into();
+    let command_output = Text::from(format!("[{}] {}", sender_name, message.0));
+
+    drop(name);
 
     ctx.game.handle(
         &mut ctx.world,
         ChatEvent {
-            message: command_output,
+            message: command_output.into(),
             position: ChatPosition::Chat,
         },
     );
--- a/server/commands/src/impls.rs
+++ b/server/commands/src/impls.rs
@@ -7,7 +7,9 @@ use crate::{
 };
 use feather_core::text::{Text, TextComponentBuilder};
 use feather_core::util::{Gamemode, Position};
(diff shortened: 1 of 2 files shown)


server/commands/src/impls.rs
```diff
@@ -186,3 +188,25 @@ pub fn whisper(
 
     Ok(())
 }
+
+#[command(usage = "say <message>")]
+pub fn say(ctx: &mut CommandCtx, message: TextArgument) -> anyhow::Result<()> {
+    let sender_name = if let Some(sender_name) = ctx.world.try_get::<Name>(ctx.sender) {
```
caelunshun, line 194: `sender_name` could be an `&str`; there's not need to clone it from what I see.

Defman: To add to this, you can also use `Option::as_ref` combined with `Option::unwrap_or_default`.

amberkowalski: > 
> 
> `sender_name` could be an `&str`; there's not need to clone it from what I see.

If you could show me what you would do to do this I would appreciate it. I've tried a few things and, other than placing the message generation within the `if let` statement, the borrow checker doesn't seem to appreciate not cloning the `String`.

caelunshun: Sure thing:
```rust
let sender_name = ctx.world.try_get::<Name>(ctx.sender).map(|name| &name.0).unwrap_or("Server");
```
Does this work?

caelunshun: Hmm, I see what the problem is... unfortunately `legion` has no `Ref::map` to fix this. I suppose performing the allocation is fine, for now.

caelunshun: Alternatively, here's a hack:

```rust
let name = world.try_get::<Name>(ctx.sender);

let sender_name = if let Some(name) = name {
    &name.0
} else {
    None
};
```

amberkowalski: I also came up with `let sender_name = ctx.world.try_get::<Name>(ctx.sender).as_deref().unwrap_or(&Name(String::from("Server"))).0.clone();` as a "more succinct" way of handling the situation (it still clones the string, however)

Defman:
```rust
let name = Some(Name("player".to_owned()));
let name: &str = name.as_ref().map(|name| &*name.0).unwrap_or("server");
```

Defman: Alternatively you can do
```rust
let name = Some(Name("player".to_owned()));
let name: &str = name.as_ref().map_or("server", |Name(n)| n);
```

amberkowalski: In every attempt to use &str I've tried, `ctx.world` gets borrowed for the lifetime of the `&str` and therefore I cant borrow it to `ctx.game.handle`

@Defman's suggestion seems to work for using the default name, but I can't use it because of the borrowing issue. 

[Here ](https://hastebin.com/vexopolulo.coffeescript) is the current code that I am using, I have indicated the spots where the borrow errors are occurring. Right now the only way I can see to avoid this is the clone.

caelunshun: With my approach above, you can add `drop(name)` before the call to `handle` and everything should work fine.

caelunshun:
Thank you!

PR #233, opened 2020-05-19
Fix Gamemode parser

amberkowalski:
The gamemode parser was using the wrong logic for its `satisfies` member, and therefore didn't work.


caelunshun:
Thanks!

Issue #228, opened 2020-05-18
Tracking issue: rigorous survival mode inventory interaction


caelunshun:
Implement `Click Window` packet actions:
* [x] `single_click`
* [x] `double_click`
* [x] `shift_click`
* [ ] `number_key`
* [ ] `item_drop`
* [x] `paint`
* [x] Write rigorous test suite to avoid duplication exploits

PR #234, opened 2020-05-19
Add number key handler for Click Window packets

amberkowalski:
Adds `number_key` handler for #228

### DIFF
--- a/server/player/src/packet_handlers/inventory.rs
+++ b/server/player/src/packet_handlers/inventory.rs
@@ -432,27 +432,39 @@ fn handle_number_key(
     packet: ClickWindow,
     key: u8,
 ) -> anyhow::Result<()> {
-    let slot = packet.slot;
+    let slot: usize = packet.slot as usize;
     let window = world.get::<Window>(player);
     let accessor = window.accessor(world)?;
 
-    if let Some(hotbar_slot_stack) = accessor.remove_item_at((35 + key) as usize)? {
+    // The slot index used to index the hotbar
+    let hotbar_slot = SlotIndex {
+        area: Area::Hotbar,
+        slot: (key - 1) as usize,
+    };
+
+    // Convert the slot index into the network slot
+    let hotbar_slot = window
+        .convert_slot(hotbar_slot, player)
+        .ok_or_else(|| anyhow::anyhow!("invalid slot index"))?;
+
+    // Perform the swap
+    if let Some(hotbar_slot_stack) = accessor.remove_item_at(hotbar_slot)? {
         // Handles the case where there is an item in both target slots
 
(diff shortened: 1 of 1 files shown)


amberkowalski:
Marked my pr as WIP because it wont work in all situations where an inventory is involved, I'm working on fixing that.

amberkowalski:
Removed WIP status since it works now.

server/player/src/packet_handlers/inventory.rs
```diff
@@ -426,13 +426,66 @@ fn handle_shift_click(
...
+    let window = world.get::<Window>(player);
+    let accessor = window.accessor(world)?;
+
+    // The slot index used to index the hotbar
+    let hotbar_slot = SlotIndex {
+        area: Area::Hotbar,
+        slot: (key - 1) as usize,
+    };
+
+    // Convert the slot index into the network slot
```
caelunshun, line 445: The `accessor` is only used to convert between indices in the protocol and internal indices (`SlotIndex`). Rather than converting `hotbar_slot` to a network slot, you can just call `inventory.item_at` directly with the `Area` and the index within the hotbar.

`WindowAccessor`'s sole purpose is when you have an index from a packet and need to index into the inventory. Otherwise, the APIs directly on `Inventory` should be used.

server/player/src/packet_handlers/inventory.rs
```diff
@@ -426,13 +426,66 @@ fn handle_shift_click(
...
+    drop(accessor);
+
+    // Create a list of slots that were updated with this operation so they can be sent to handlers
+    let slots = smallvec![
+        window
+            .convert_network(slot as usize)
+            .ok_or_else(|| anyhow::anyhow!("invalid slot index"))?
+            .into(),
+        window
+            .convert_network((35 + key) as usize)
```
caelunshun, line 480: These can also be listed as `SlotIndex { area: Area::Hotbar, key }`; there's no need to do an intermediate conversion.

server/player/src/packet_handlers/inventory.rs
amberkowalski: How would I get the `Inventory` in this case? Sorry if I'm missing something obvious.

server/player/src/packet_handlers/inventory.rs
amberkowalski: Thanks!

server/player/src/packet_handlers/inventory.rs
caelunshun: `Inventory` is a component, just like `Window`—just call `World::get::<Inventory>` :)

caelunshun:
Thanks!

PR #241, opened 2020-05-21
Full block interaction

amberkowalski:
Placement:  
- [x]  Handle placing blocks with orientation
- [x]  Handle cases where the target block should be overridden by the placed block (stone -> grass)
- [x] ~~Handle waterlogged blocks~~
- [x] ~~Shift + Right-Click~~
Interaction:
- [x] Open the correct windows on the client when they interact with a block that has one (added interaction handler trait and hashmap for adding interaction handlers)
Handling the inventories / buttons in the windows will be handled in a separate pr.

Thanks to @ssttevee for the help!


server/player/src/packet_handlers/placement.rs
```diff
@@ -24,66 +24,102 @@ pub fn handle_player_block_placement(
...
                 }
             };
 
-            // TODO: waterlogged blocks, more
-            let pos = match placed_on.kind() {
-                BlockKind::Grass | BlockKind::TallGrass | BlockKind::Water | BlockKind::Lava => {
-                    packet.location
+            // Decide whether the player should place a block or interact with the block they are targeting
+            // TODO: Maybe player shifting may need to be taken into account (shift click on interactable block)
+            match target_block.kind() {
```
amberkowalski, line 41: It would be cool if feather-data was able to serialize whether the block can be interacted with, but I'm not sure how feasible that is.

caelunshun: Not possible, since vanilla doesn't give us that information.

What you can do is manually write the mappings in `feather-definitions`. Take a look at the `data` directory there; it has RON files which specify e.g. block properties. You can add a boolean property "interactable" there.

server/player/src/packet_handlers/placement.rs
```diff
@@ -24,66 +24,102 @@ pub fn handle_player_block_placement(
...
+
+/// Handles interaction with blocks that have an inventory
+pub fn handle_block_interaction(
+    game: &mut Game,
+    world: &mut World,
+    player: Entity,
+    target_block_kind: BlockKind,
+    packet: PlayerBlockPlacement,
+) {
+    let gamemode = *world.get::<Gamemode>(player);
```
caelunshun, line 124: It might be a good idea to have a global `HashMap<BlockKind, Box<dyn InteractionHandler>>` to handle interaction. Otherwise, this might get messy.

amberkowalski: Would also get rid of the `Interactable` property requirement

server/player/src/packet_handlers/placement.rs
```diff
@@ -24,66 +40,105 @@ pub fn handle_player_block_placement(
...
+
+    let inventory = world.get::<Inventory>(player);
+
+    let item = match inventory.item_in_main_hand(player, world) {
+        Some(item) => item,
+        None => return, // No block to place
+    };
+
+    drop(inventory);
```
Defman, line 89: We can avoid drops by scoping
```rust
let item = {
  let inventory = world.get::<Inventory>(player);
  match inventory.item_in_main_hand(player, world) {
    Some(item) => item,
    // Offhand?
    None => return, // No block to place
  };
};
```

server/player/src/packet_handlers/placement.rs
```diff
@@ -24,66 +40,105 @@ pub fn handle_player_block_placement(
...
+        let item = ItemStack::new(item.ty, item.amount - 1);
+        inventory
+            .set_item_at(Area::Hotbar, held_item, item)
+            .unwrap();
+
+        let event = InventoryUpdateEvent {
+            slots: smallvec![slot(Area::Hotbar, held_item)],
+            player,
+        };
+        drop(inventory);
```
Defman, line 130: Use scoping to update the inventory.

server/player/src/packet_handlers/block_interaction/crafting_table.rs
```diff
@@ -0,0 +1,33 @@
...
+        &self,
+        _game: &mut Game,
+        world: &mut World,
+        player: Entity,
+        window_id: u8,
+    ) {
+        // Show the crafting table GUI for the player
+        let open_window_packet = Box::new(OpenWindow {
+            window_id,
+            window_type: String::from("minecraft:crafting_table"),
```
caelunshun, line 21: The type of `window_type` should probably be changed to `&'static str`, since we don't require dynamic strings here.

server/player/src/packet_handlers/block_interaction/crafting_table.rs
```diff
@@ -0,0 +1,33 @@
...
+            window_title: TextRoot::from("Crafting Table").into(),
+            number_of_slots: 0,
+            entity_id: None,
+        });
+
+        let network = world.get::<Network>(player);
+
+        network.send_boxed(open_window_packet);
```
caelunshun, line 29: Call `network.send(open_window_packet)` and don't box the packet (`send_boxed` is only required when you have a `Box<dyn Packet>` and don't know the actual type.)

server/player/src/packet_handlers/placement.rs
```diff
@@ -24,66 +42,102 @@ pub fn handle_player_block_placement(
...
 
-            // TODO: waterlogged blocks, more
-            let pos = match placed_on.kind() {
-                BlockKind::Grass | BlockKind::TallGrass | BlockKind::Water | BlockKind::Lava => {
-                    packet.location
-                }
-                _ => packet.location + packet.face.placement_offset(),
-            };
+            // Decide whether the player should place a block or interact with the block they are targeting
+            // TODO: Maybe player shifting may need to be taken into account (shift click on interactable block)
```
caelunshun, line 58: This is something to do after sneaking gets implemented. Currently, the server isn't aware of it.

server/types/src/components.rs
```diff
@@ -104,3 +104,16 @@ impl MessageReceiver {
         self.buffer.drain(..)
     }
 }
+
+/// Component which stores a number which used for OpenWindow packets, incremented on every access
+#[derive(Default, Debug)]
+pub struct OpenWindowCount {
```
caelunshun, line 110: I believe the counter is supposed to start at 1, according to the wiki page? `u8::default()` returns 0

server/player/src/packet_handlers/placement.rs
amberkowalski: Oh hahaha, thanks!

server/player/src/packet_handlers/block_interaction/crafting_table.rs
amberkowalski: Okay, thanks!

server/types/src/components.rs
amberkowalski: I saw no issues with it starting at 0 but I can change it to follow the wiki.

server/player/src/packet_handlers/block_interaction/crafting_table.rs
amberkowalski: That causes issues with `Packet`'s `read_from` method, since I cant `try_get_str` and even if I could, it wouldn't be static

core/network/src/packets.rs
```diff
@@ -1484,13 +1537,56 @@ pub struct ConfirmTransactionClientbound {
...
+}
+
+impl Packet for OpenWindow {
+    fn read_from(&mut self, buf: &mut Cursor<&[u8]>) -> anyhow::Result<()> {
+        self.window_id = buf.try_get_u8()?;
+        self.window_type = buf.try_get_string()?;
+        self.window_title = buf.try_get_string()?;
+        self.number_of_slots = buf.try_get_u8()?;
+
+        if self.window_type == "EntityHorse" {
```
caelunshun, line 1556: This could be written more concisely with

```rust
self.entity_id = if self.window_type == "EntityHorse" { Some(buf.try_get_i32()? } else { None };
```

caelunshun:
Thank you!

PR #242, opened 2020-05-22
Add /me command

onbjerg:
Adds `/me <action>` command for #229 which is described [here](https://minecraft.gamepedia.com/Commands/me).

This has not been tested on multiple accounts. The text for the command was deduced from running `/me` on a 1.15.2 server from both the console (which appears as `@`) and a normal player account.

### DIFF
--- a/server/commands/src/impls.rs
+++ b/server/commands/src/impls.rs
@@ -215,13 +215,12 @@ pub fn say(ctx: &mut CommandCtx, message: TextArgument) -> anyhow::Result<()> {
 }
 
 #[command(usage = "me <action>")]
-pub fn me(ctx: &mut CommandCtx, action: TextArgument) -> anyhow::Result<()> {
-    let name = ctx.world.try_get::<Name>(ctx.sender);
-
-    let sender_name = if let Some(name) = &name { &name.0 } else { "@" };
-    let command_output = Text::from(format!("* {} {}", sender_name, action.0));
-
-    drop(name);
+pub fn me(ctx: &mut CommandCtx, TextArgument(action): TextArgument) -> anyhow::Result<()> {
+    let command_output = {
+        let name = ctx.world.try_get::<Name>(ctx.sender);
+        let sender_name = name.as_ref().map_or("@", |Name(n)| n);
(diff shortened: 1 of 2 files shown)


server/commands/src/impls.rs
```diff
@@ -213,3 +213,23 @@ pub fn say(ctx: &mut CommandCtx, message: TextArgument) -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[command(usage = "me <action>")]
+pub fn me(ctx: &mut CommandCtx, action: TextArgument) -> anyhow::Result<()> {
+    let name = ctx.world.try_get::<Name>(ctx.sender);
+
+    let sender_name = if let Some(name) = &name { &name.0 } else { "@" };
```
Defman, line 221:
```rust
let command_output = {
  let name = ctx.world.try_get::<Name>(ctx.sender);
  let sender_name = name.as_ref().map_or("server", |Name(n)| n);
  Text::from(format!("* {} {}", sender_name, action.0));
};
```

server/commands/src/impls.rs
```diff
@@ -213,3 +213,23 @@ pub fn say(ctx: &mut CommandCtx, message: TextArgument) -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[command(usage = "me <action>")]
+pub fn me(ctx: &mut CommandCtx, action: TextArgument) -> anyhow::Result<()> {
```
Defman, line 218: You can use pattern matching in the function signatur for `TextArgument` instead of doing `action.0`.

```rust
fn me(ctx: &mut CommandCtx, TextArgument(action): TextArgument) -> String {
    name
}
```

Defman: Perhaps it would be better to implement `AsRef<str>` for `TextArgument`.

onbjerg: Ooh, that's clever, I didn't know you could do that! Thanks

server/commands/src/impls.rs
onbjerg: Much neater, thanks!

caelunshun:
Thank you!

PR #243, opened 2020-05-22
Add notice about JDK 8 requirement for compiling

onbjerg:
To clear up some initial confusion I had when first building from source


caelunshun:
Thanks!

PR #247, opened 2020-05-23
Fix falling blocks

ssttevee:
Before:

After:


caelunshun:
Whoops, missed that regression after loot tables were implemented. Thank you!

PR #248, opened 2020-05-23
Handle placing blocks with cardinality

ssttevee:
The logic for determining cardinality feels like it doesn't belong there. Where do you think it should live?

Also, for the code in `categories.rs`, should those be part of the generated code? I could extend the generator to generate `has_x` functions.

Side note: I'd like to work towards working redstone; should I submit many small PRs or fewer big ones?

### DIFF
--- a/core/blocks/build.rs
+++ b/core/blocks/build.rs
@@ -11,12 +11,14 @@ fn main() {
             let _ = std::fs::create_dir_all(base);
 
             let block_fns = format!("{}/block_fns.rs", base);
+            let props = format!("{}/properties.rs", base);
             let table = format!("{}/table.rs", base);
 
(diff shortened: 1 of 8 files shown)


ssttevee:
I just realized #241 addresses this 👀

amberkowalski:
I can merge this into the pr with your permission, since I have changed how block placement works programmatically but haven't implemented cardinality.

ssttevee:
That sounds good, but I think I should include all the cardinalities (down + cubic) first as well as figure out the situation with the categories.

amberkowalski:
If you want I think you can open prs on my development fork and I can merge them in whenever you get that done, thanks for working on this :)

ssttevee:
merged into #241

PR #252, opened 2020-05-23
Gamerules

Defman:
Do we want to store gamerules as just fields or do want to store them as a kind of `Set` or `&[u8]`?


InfinityZ25:
Store them as sets. It's neater and the difference in memory is not too great.

Defman:
I dropped the idea of using new types and instead used a single struct with fields as the rules.

caelunshun:
@Defman is this ready for merging?

Defman:
Yes and no, missing serialization and deserialization.

Defman:
@caelunshun Lets merge it and ill submit a separate PR for saving and loading.

caelunshun (approved):

PR #260, opened 2020-06-07
Implement /stop command

Deliquescence:
For #229

I also added the `TextValue::translate` fn to use for translate keys which don't need replacement, like "stopping server" or "server closed".

Of course, let me know if there could be improvements anywhere :)


caelunshun:
Thanks!

PR #261, opened 2020-06-07
Supported blocks (Issue #135)

### LINKED ISSUE #135: Break non-solid blocks when the block underneath is destroyed

Schuwi:
Some blocks in Minecraft can't exist without others. Grass needs a block beneath it, torches need a block beneath or besides them to provide support. There are even more complex conditions like sugar cane, which needs either another sugar cane block beneath it or water next to the block beneath it.

This functionality of supported blocks is implemented by the following components:
- `feather-server-util::is_block_supported_at` determines whether a block of a specific `BlockId` is supported at a specific `BlockPosition` in the world
- `feather-server-util::does_block_need_support` determines whether a block of a specific `BlockKind` has special support requirements that need to be checked with `is_block_supported_at`
- `feather-server-entity::object::supported_blocks::break_unsupported_blocks` system which works analogous to the `falling_block::spawn_falling_blocks` system and uses `BlockNotifySupportedBlock` to check notified blocks for support
- a check in `feather-server-player::packet_handlers::placement::handle_block_placement` to prevent placing down blocks which would be unsupported

Unresolved issues:
- [x] The big one: I am totally unsure about performance here. I don't know how far the compiler will be able to optimize this implementation, I need a second opinion on this
- [ ] I feel like `break_unsupported_blocks` system doesn't really belong into `feather-server-entity`, maybe move it to `feather-server-block` once #257 is implemented?
- [ ] Naming in general. Maybe re-term supported blocks to attached blocks? That's what some other implementations call it but supported blocks might actually describe this better, as for example I wouldn't call those sugar cane conditions an attachement

This would resolve #135.


Schuwi:
Regarding the performance:

I suspect as long as the compiler inlines the `BlockId::needed_support` call in `is_block_supported_at` it should be good performance-wise as the compiler should then be able to resolve the recursive function statically (??) So maybe add an `#[inline]` hint on `BlockId::needed_support` to be safe?

caelunshun:
Thanks, great work!

> * The big one: I am totally unsure about performance here. I don't know how far the compiler will be able to optimize this implementation, I need a second opinion on this

I wouldn't worry about performance here—it's not hot code, and the implementation is reasonable enough I doubt we'll have issues. Unless we find profile results to indicate otherwise, this PR should be good enough in terms of performance. 
> * I feel like `break_unsupported_blocks` system doesn't really belong into `feather-server-entity`, maybe move it to `feather-server-block` once #257 is implemented?
 
I agree; it should be moved into the `server-block` crate once the block entities PR gets merged.

> * Naming in general. Maybe re-term supported blocks to attached blocks? That's what some other implementations call it but supported blocks might actually describe this better, as for example I wouldn't call those sugar cane conditions an attachement

I think both names are reasonable—they both describe the functionality well. Going with "supported" is good so long as we keep it consistent.

Schuwi:
Okay, so with this last rewrite performance shouldn't be a concern (to me) anymore and it makes the code way more readable. I don't have a strong opinion on naming so unless somebody else disagrees with the naming I'd just leave it like that.

And unless I missed something when looking through the Minecraft block list or my game knowledge failed me this should now be a complete implementation of all blocks that require some kind of support.
I tested it as far as was possible. As stuff like water and block light are not implemented yet especially some of the more complex block types remain untested.

So from my side this would be ready for merging now.

PR #269, opened 2020-06-16
Fix various world loading problems (Issue #265)

### LINKED ISSUE #265: Can't load vanilla world
According to the README, loading a vanilla save should work but it doesn't.

    2020-06-10 23:53:53,133 WARN  [feather_server_chunk::chunk_manager] Failed to load chunk at ChunkPosition { x: 3, z: -3 }: Region file contains invalid NBT: invalid type: map, expected a sequence

This is the error feather produces for basically every chunk the server wants to load.
Investigating a bit further it seems that the NBT chunk format implementation is wrong.
`feather-anvil::region::ChunkLevel.heightmaps` is a Vec<i64> but inspecting the vanilla save in an NBT viewer `Level/Heightmaps` is actually a struct of multiple [u64; 36] (basically [u9; 256] packed into u64 array) fields. (And also a bit different field names than those listed on the Minecraft Wiki.)

When temporarily circumventing this error I also get a plethora of other errors:

    2020-06-11 01:33:02,281 WARN  [feather_server_chunk::chunk_manager] Failed to load chunk at ChunkPosition { x: -9, z: -15 }: Chunk contains invalid block minecraft:vine
    2020-06-11 01:33:02,281 WARN  [feather_server_chunk::chunk_manager] Failed to load chunk at ChunkPosition { x: -9, z: -14 }: Chunk contains invalid block minecraft:lava
    2020-06-11 01:33:02,330 WARN  [feather_server_chunk::chunk_manager] Failed to load chunk at ChunkPosition { x: -9, z: -7 }: Chunk contains invalid block minecraft:water
    2020-06-11 01:33:02,330 WARN  [feather_server_chunk::chunk_manager] Failed to load chunk at ChunkPosition { x: -8, z: -13 }: Region file contains invalid NBT: invalid type: integer `5`, expected f32
    2020-06-11 01:33:02,380 WARN  [feather_server_chunk::chunk_manager] Failed to load chunk at ChunkPosition { x: -8, z: -7 }: Chunk contains invalid block minecraft:oak_log
    2020-06-11 01:33:02,380 WARN  [feather_server_chunk::chunk_manager] Failed to load chunk at ChunkPosition { x: -7, z: -5 }: Chunk contains invalid block minecraft:birch_log
    2020-06-11 01:33:02,480 WARN  [feather_server_chunk::chunk_manager] Failed to load chunk at ChunkPosition { x: -2, z: -14 }: Chunk contains invalid block minecraft:oak_fence
    2020-06-11 01:33:02,480 WARN  [feather_server_chunk::chunk_manager] Failed to load chunk at ChunkPosition { x: -2, z: -11 }: Chunk contains invalid block minecraft:tall_grass
    2020-06-11 01:33:02,530 WARN  [feather_server_chunk::chunk_manager] Failed to load chunk at ChunkPosition { x: 3, z: -15 }: Chunk contains invalid block minecraft:rail

(There were many of each of those errors but I picked only one of each kind for the issue.)
I haven't investigated those further yet.

I'm working on fixing the heightmaps implementation. Shouldn't be that hard, just need to implement a de/serializer for those packed unsigned 9-bit arrays.

Schuwi:
This PR aims to fix a few issues I encoutered when trying to load a vanilla world. Mainly this fixes the format of heightmaps in chunk data.
This is basically ready to merge, just needs a little polish. Namely:
- [X] Remove panics from `packed_u9`.
    (I just can't figure out how to return an error from `serialize` or `deserialize`. I thought `serde::ser::Error::custom("my error")` would do the trick but rustc says nope.)
- [X] Complete the test implementation for `packed_u9`/`Heightmaps::packed_i64_vec`.
- [X] Run `cargo fmt` and check `cargo clippy`.

Resolves #265.

### DIFF
--- a/core/anvil/src/entity.rs
+++ b/core/anvil/src/entity.rs
@@ -131,8 +131,6 @@ pub struct BaseEntityData {
     pub rotation: ArrayVec<[f32; 2]>,
     #[serde(rename = "Motion")]
     pub velocity: ArrayVec<[f64; 3]>,
-    #[serde(rename = "Health")]
-    pub health: f32,
 }
 
 impl BaseEntityData {
@@ -149,7 +147,6 @@ impl BaseEntityData {
             String::from("Motion"),
             Value::List(self.velocity.iter().copied().map(Value::Double).collect()),
         );
-        map.insert(String::from("Health"), Value::Float(self.health));
     }
 }
 
@@ -161,12 +158,11 @@ pub enum EntityLoadError {
 
 impl BaseEntityData {
     /// Creates a `BaseEntityData` from its parameters.
-    pub fn new(pos: Position, velocity: Vec3d, health: f32) -> Self {
+    pub fn new(pos: Position, velocity: Vec3d) -> Self {
         Self {
             position: [pos.x, pos.y, pos.z].into(),
             rotation: [pos.yaw, pos.pitch].into(),
             velocity: [velocity.x, velocity.y, velocity.z].into(),
-            health,
         }
     }
 
@@ -202,7 +198,6 @@ impl Default for BaseEntityData {
             position: [0.0, 0.0, 0.0].into(),
             rotation: [0.0, 0.0].into(),
             velocity: [0.0, 0.0, 0.0].into(),
-            health: 20.0,
         }
     }
 }
@@ -211,11 +206,33 @@ impl Default for BaseEntityData {
 pub struct AnimalData {
     #[serde(flatten)]
     pub base: BaseEntityData,
+    #[serde(rename = "Health")]
+    pub health: f32,
 }
 
 impl AnimalData {
     fn write_to_map(self, map: &mut HashMap<String, Value>) {
         self.base.write_to_map(map);
+        map.insert(String::from("Health"), Value::Float(self.health));
+    }
+}
+
+impl AnimalData {
+    /// Creates a `BaseEntityData` from its parameters.
+    pub fn new(base: BaseEntityData, health: f32) -> Self {
+        Self {
+            base,
+            health
+        }
+    }
+}
+
+impl Default for AnimalData {
+    fn default() -> Self {
+        AnimalData {
+            base: Default::default(),
+            health: 20.0,
+        }
     }
 }
 
@@ -223,14 +240,14 @@ impl AnimalData {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ItemData {
     #[serde(rename = "Count")]
-    pub count: u8,
+    pub count: i8,
     #[serde(rename = "id")]
     pub item: String,
 }
 
 impl ItemData {
     fn write_to_map(self, map: &mut HashMap<String, Value>) {
-        map.insert(String::from("Count"), Value::Byte(self.count as i8));
+        map.insert(String::from("Count"), Value::Byte(self.count));
         map.insert(String::from("id"), Value::String(self.item));
     }
 }
@@ -255,9 +272,11 @@ pub struct ItemEntityData {
     #[serde(rename = "Age")]
     pub age: i16,
     #[serde(rename = "PickupDelay")]
-    pub pickup_delay: u8,
+    pub pickup_delay: i16,
     #[serde(rename = "Item")]
     pub item: ItemData,
+    #[serde(rename = "Health")]
+    pub health: i16,
 }
 
 impl ItemEntityData {
@@ -271,8 +290,9 @@ impl ItemEntityData {
         map.insert(String::from("Age"), Value::Short(self.age));
         map.insert(
             String::from("PickupDelay"),
-            Value::Byte(self.pickup_delay as i8),
+            Value::Short(self.pickup_delay),
         );
+        map.insert(String::from("Health"), Value::Short(self.health));
     }
 }
 
@@ -288,14 +308,14 @@ pub struct ArrowEntityData {
     // TODO: Change this field to `bool` when issue with hematite_nbt is resolved.
     // See: https://github.com/PistonDevelopers/hematite_nbt/issues/43
     #[serde(rename = "crit")]
-    pub critical: u8,
+    pub critical: i8,
 }
 
 impl ArrowEntityData {
     fn write_to_map(self, map: &mut HashMap<String, Value>) {
         self.entity.write_to_map(map);
 
-        map.insert(String::from("crit"), Value::Byte(self.critical as i8));
+        map.insert(String::from("crit"), Value::Byte(self.critical));
     }
 }
 
@@ -310,7 +330,6 @@ mod tests {
             position: [1.0, 2.0, 3.0].into(),
             rotation: [4.0, 5.0].into(),
             velocity: [6.0, 7.0, 8.0].into(),
-            health: 20.0,
         };
         let pos = data.read_position().unwrap();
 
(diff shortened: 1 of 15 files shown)


Schuwi:
I spent waaaay to much time on those packed u9 heightmap arrays. Finally everything works, tests are implemented and deserializing a vanilla chunk gives sensible heightmap data.
This should now be ready for merge and allow for some sweet vanilla world loading once again.

Schuwi:
Huh, I just can't replicate that clippy warning. Even running explicitely with `cargo clippy -- -D clippy::inconsistent-digit-grouping` executes without warnings. I'll just add a `#[allow(clippy::inconsistent-digit-grouping)]` because the digit grouping makes sense here.

core/anvil/src/entity.rs
```diff
@@ -211,26 +206,45 @@ impl Default for BaseEntityData {
...
 
 impl AnimalData {
     fn write_to_map(self, map: &mut HashMap<String, Value>) {
         self.base.write_to_map(map);
+        map.insert(String::from("Health"), Value::Float(self.health));
+    }
+}
+
+impl AnimalData {
+    /// Creates a `BaseEntityData` from its parameters.
```
caelunshun, line 221: Looks like `BaseEntityData` should be changed to `AnimalData` in this comment

core/anvil/src/region/mod.rs
```diff
@@ -31,62 +31,232 @@ const DATA_VERSION: i32 = 1631;
...
+    motion_blocking_no_leaves: Vec<u16>,
+    #[serde(with = "packed_u9", default)]
+    ocean_floor: Vec<u16>,
+    #[serde(with = "packed_u9", default)]
+    world_surface: Vec<u16>,
+}
+
+impl Heightmaps {
+    // TODO remove this and use hematite-nbt serde for serialization
+    pub(crate) fn pack_u9(u9_array: Vec<u16>) -> Vec<i64> {
```
caelunshun, line 77: It looks like taking `u9_array: &[u16]` should work here—there's no need for an owned `Vec`.

core/anvil/src/region/mod.rs
```diff
@@ -31,62 +31,232 @@ const DATA_VERSION: i32 = 1631;
...
+        let mut packed = Vec::with_capacity(36);
```
caelunshun, line 78: Could the initial capacity be derived from the length of the input, instead of using this magic number?  (Or at least extract it into a constant, since it's used a few more times in this file.)

core/anvil/src/region/mod.rs
```diff
@@ -31,62 +31,232 @@ const DATA_VERSION: i32 = 1631;
...
+        let mut iter = u9_array.iter();
+
+        let mut container = 0u64;
+        let mut shift = 0;
+        for _elem in 0..256 {
+            // For every element (u9)
+            let element = *iter.next().unwrap() as u64;
```
caelunshun, line 85: We should return a `Result::Err` here rather than panicking. Right now, if we load in invalid heightmap data and then that chunk is saved, the server will panic. I'd prefer the chunk save process to fail with a log message.

core/anvil/src/region/mod.rs
```diff
@@ -31,62 +31,232 @@ const DATA_VERSION: i32 = 1631;
...
+                    return Err(A::Error::custom("sequence length not equal to 36"));
+                }
+
+                let mut u9_array: Vec<u16> = Vec::with_capacity(256);
+
+                let mut container: Option<u64> = seq.next_element()?.map(|x: i64| x as u64); // We checked the length
+                let mut shift = 0;
+                for _elem in 0..256 {
+                    // For every element (u9)
+                    let mut element: u16 = ((container.unwrap() >> shift) & 0x1FF) as u16;
```
caelunshun, line 156: Same panicking problem applies here—rather than unwrapping, I'd prefer to return an error.

core/anvil/src/region/mod.rs
```diff
@@ -31,62 +31,232 @@ const DATA_VERSION: i32 = 1631;
...
+                    shift += 9;
+
+                    if shift >= 64 {
+                        // Take next container
+                        container = seq.next_element()?.map(|x: i64| x as u64);
+
+                        element |= if shift > 64 {
```
caelunshun, line 163: I think

```rust
if shift > 64 {
    element |= container.unwrap() << -(shift - 64 - 9) & 0x1FF;
}
```
is equivalent here, and the intent is more clear.

(Rather than bitwise ORing with 0 if the condition is false.)

core/chunk/src/lib.rs
```diff
@@ -244,31 +257,85 @@ impl Chunk {
...
+            HeightMapMask,
+            &'static dyn Fn(&HeightMap) -> u16,
+            &'static dyn Fn(&mut HeightMap, u16) -> (),
+        );
 
-        if block.is_solid() && heightmap.ocean_floor() < y as u8 {
-            heightmap.set_ocean_floor(y as u8);
-            mask |= HeightMapMask::OCEAN_FLOOR;
-        }
+        let checks: Vec<HeightMapCheckContext> = vec![
```
caelunshun, line 278: I'm not so sure about this allocation—this is potentially very hot code. It looks to me that using a fixed array instead of a `Vec` should work here.

Schuwi:
Thanks for all the comments! I'm sorry there had to be so many. I got a bit sloppy in the end and just wanted to be done with it. This turned into a way bigger thing than I thought it would.
But now that https://github.com/PistonDevelopers/hematite_nbt/pull/52 is merged I can fully concentrate on this and I'll take a look at your comments tomorrow with a clear head :)

core/anvil/src/region/mod.rs
Schuwi: Yeah, I didn't put much effort into this method as I thought it to be temporary but I'll give it a once-over. Actually the server panics when it encounters a serialization error on saving regardless right now, because the Result is just being `expect`ed in the chunk save function

core/anvil/src/region/mod.rs
Schuwi: Interesting that clippy doesn't catch stuff like that

core/chunk/src/lib.rs
Schuwi: Well, I mean optimally this whole vec should be optimized out and the loop unrolled with the functions inlined, I just did it like this to avoid redundant code. I'll change it to an array but I doubt it'll make a difference after optimization.

core/anvil/src/region/mod.rs
Schuwi: Honestly, I'd rather just leave the panic in for now and remove this whole `Blob` business instead in the next PR. Switching everything to use the serde derived `Serialize` instead, now that array serialization is supported. Because right now everything in the chunk saving pipeline just panics when it encounters some error and I don't see the need to implement proper error propagation and handling just to scrap all of it in the next PR.

core/anvil/src/region/mod.rs
Schuwi: This unwrapping is actually alright, this would only panic if there is a programming error in the algorithm or in the seqaccess

Schuwi:
Alright, so I have actually implemented serde serialization for chunk data in my local working copy now and everything works. I am not sure whether to include it in this PR or put it into a separate one. That would also fix the server crashing when failing to save a chunk and put out a warning instead.

caelunshun:
> Alright, so I have actually implemented serde serialization for chunk data in my local working copy now and everything works. I am not sure whether to include it in this PR or put it into a separate one. That would also fix the server panicking when failing to save a chunk and put out a warning instead.

Nice, thanks for that! I think the new serialization probably belongs in another PR—this PR is pretty much ready to merge.

Schuwi:
Alright, then everything else would be for another PR so this would be ready to merge from my side.

caelunshun:
Great, thank you!

PR #271, opened 2020-06-19
Redstone

Inky-developer:
Implements basic redstone behaviour.

Implemented:
- [x]  Redstone wire
- [x]  Redstone block
- [x]  Redstone torch
- [ ]  Repeater
- [ ]  Buttons, Levers, Pressure plates & Tripwire hook
- [ ]  Doors, Trapdoors & Fence gates
- [ ]  Observer
- [ ]  Comparator
- [ ]  Redstone lamp
- [ ]  Daylight sensor
- [ ]  Piston & Sticky Piston
- [ ]  Note block
- [ ]  Dropper & Dispenser
- [ ]  Hopper
- [ ]  Trapped chest
- [ ] Tnt

This is my first contribution on github, so if I did something wrong please let me know :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1138,6 +1138,7 @@ dependencies = [
  "ahash",
  "anyhow",
  "feather-core",
+ "feather-server-block",
  "feather-server-commands",
  "feather-server-entity",
  "feather-server-network",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1247,6 +1247,7 @@ version = "0.6.0"
 dependencies = [
  "crossbeam",
  "feather-core",
+ "feather-server-block",
  "feather-server-chunk",
  "feather-server-network",
  "feather-server-player",
--- a/core/blocks/src/lib.rs
+++ b/core/blocks/src/lib.rs
@@ -9,8 +9,8 @@ mod directions;
 #[allow(warnings)]
 #[allow(clippy::all)]
 mod generated;
-mod wall_blocks;
 mod redstone_blocks;
+mod wall_blocks;
 
 static BLOCK_TABLE: Lazy<BlockTable> = Lazy::new(|| {
     let bytes = include_bytes!("generated/table.dat");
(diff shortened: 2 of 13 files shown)


Pospelove:
Good luck

server/types/src/game.rs
```diff
@@ -56,6 +57,10 @@ pub struct Game {
     pub time: Time,
     /// The event handler map.
     pub event_handlers: Arc<EventHandlers>,
+    // The event scheduler
+    // ToDo: If this struct is accessible from multiple threads,
+    // shouldn't this use a Mutex?
+    pub event_scheduler: EventScheduler,
```
Inky-developer, line 63: What would be the best way to implement such an event scheduler?

Defman: For timing redstone, cloks and etc? If so, you might want to implement tile ticks instead and just use a system to advance the tile ticks.

Inky-developer: That is basically what I have implemented here. What I was wondering is if something like a mutext is neccessary here. Sorry for the confusion, as I am currently learing rust I am not sure what the appropriate way is to store this event scheduler.

Schuwi: It would probably best to tie those tile ticks in with the current block update system used for creating falling blocks and the block support checks right?
When a block is changed we notify the neighbouring blocks by creating block update entities which are handled by other systems to actually do the required work. I think it would make sense to add a property to those entities to track in how many ticks or at what specific tick they should be processed and only act on them in the corresponding system when that time is reached.

Defman: Ahh ok, then it should be fine and no need for Mutex; take a look at how chunk entities are handled, they are very similar.

ColdIce1605:
Is dead?

Inky-developer:
Honestly, my code is pretty terrible. So if I continued, I'd redo the code from scratch.
Also, I am not certain how exactly the future of feather is planned. I heard that it might be planned to move all the behaviour into plugins in which case this pull request would be obsolete.

I am interested in continuing my work if someone (eg. @caelunshun) thinks that this pr is still worth it.

PR #272, opened 2020-06-23
Remove chunk data serialization workaround in feather-anvil

Schuwi:
Because serde derived serialization of the NBT Byte/Int/LongArray type was previously impossible via `hematite_nbt`, a workaround was in place, implementing serialization manually via the non-serde `hematite_nbt` API.
Now that serde serialization support is implemented via https://github.com/PistonDevelopers/hematite_nbt/pull/52 this PR seeks to remove the manual serialization and use serde derive macros in `feather-anvil` instead. This removes a lot of verbose code and makes improving the data model way faster and less error-prone.

I tested loading, flying around in and saving of a vanilla world. No errors occured and everything seemed to be fine though I haven't inspected the NBT files. All tests complete successfully.

This ties in with my work on world loading problems in #269.


core/anvil/src/serialization_helper.rs
```diff
@@ -0,0 +1,202 @@
...
+        nbt::i64_array(packed_iter, serializer)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use core::iter;
+    use serde::{Deserialize, Serialize};
+    use serde_test::Token;
```
Schuwi, line 117: Well, not sure what Clippy is on but this import is certainly not unused. Removing it just causes the tests compilation to fail

Schuwi:
Weeeell, today I've learnt to be nice to clippy, and run it with `--all-targets`. It... tends to help :sweat_smile:

Issue #205, opened 2020-04-29
Item duplication


besuper:
If you drop items when your inventory is open your item going to be duplicate by 2

![duplication](https://user-images.githubusercontent.com/16976439/80652737-e3c0e680-8a78-11ea-971c-6d5773aa2a94.gif)

caelunshun:
This would be a bug in `server/entity/object/item.rs` in the `item_collect` system. Most likely caused by our use of `unsafe` there; might be worth sacrificing a bit of performance to remove the unsafe code.

amberkowalski:
Additionally, it appears that if you close your inventory quickly after dropping an item from your inventory, the item disappears.

caelunshun:
I've tracked this down to what appears to be a client-side bug/abnormality. Contrary to my initial belief, this is not a bug in the `item_collect` system, nor does it have anything to do with unsafe code. Rather, what appears to happen is this:
* Under certain conditions (what exactly these are I have no idea), the creative mode client fails to send a Creative Inventory Action packet to update a slot where it picked up an item. This means that, when a user picks up an item and puts it on their cursor, the server still thinks that the original slot contains that same item.
* When the user clicks outside the window, the client sends Creative Inventory Action to drop the item. It is then duplicated, because on the server side, the item still occupies its original slot.

I'm not sure what we're doing differently from the vanilla server which causes this. We will need further investigation to find a solution.

Note that this is only a bug in creative mode; it will not happen in survival because Creative Inventory Action is not accepted.

AnEthernetCable:
Duplication of slabs is also an issue - when breaking in survival, a single slab drops 2 slabs. Should this be a separate issue?

caelunshun:
> Duplication of slabs is also an issue - when breaking in survival, a single slab drops 2 slabs. Should this be a separate issue?

Yes, that's a separate, unrelated issue. It arises from incomplete handling of loot table conditions.

The loot table for a slab looks e.g. like this:
```json
{
  "type": "minecraft:block",
  "pools": [
    {
      "rolls": 1,
      "entries": [
        {
          "type": "minecraft:item",
          "functions": [
            {
              "function": "minecraft:set_count",
              "conditions": [
                {
                  "condition": "minecraft:block_state_property",
                  "block": "minecraft:acacia_slab",
                  "properties": {
                    "type": "double"
                  }
                }
              ],
              "count": 2
            },
            {
              "function": "minecraft:explosion_decay"
            }
          ],
[...]

PR #273, opened 2020-06-28
Update TextArgument with fixed advance_to_end

BuggStream:
This is a follow up MR related to feather-rs/lieutenant#13. As far as I am aware this is the only instance in the code base where `advance_to_end` should be used.

NOTE: I also added the `/me` command to the `commands` macro. For some reason this command was left out. Let me know if this should be done in a separate MR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1900,8 +1900,8 @@ checksum = "c7d73b3f436185384286bd8098d17ec07c9a7d2388a6599f824d8502b529702a"
 
 [[package]]
 name = "lieutenant"
-version = "0.1.0"
-source = "git+https://github.com/feather-rs/lieutenant#52a14a6e30bd3d6335d1883fcdc8d36cc89dbae0"
+version = "0.2.0"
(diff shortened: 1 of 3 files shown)


BuggStream:
I was wondering if it would be useful to write tests for these `ArgumentKind` implementations? Not just for the one I changed, but for all of them.

amberkowalski:
@caelunshun looks good for merging if you want to.

caelunshun:
> I was wondering if it would be useful to write tests for these `ArgumentKind` implementations? Not just for the one I changed, but for all of them.

Yes, that would be great! Tests are definitely lacking right now. When I have time tomorrow I'll write some test cases for the command parsers.

(On a related note, I've got a refactor going which should make it a lot easier to write tests for gameplay-related functions—hopefully that can improve reliability overall.)

amberkowalski:
Oh does this mean writing duplication bug tests will be possible(/easier)?

caelunshun:
> Oh does this mean writing duplication bug tests will be possible(/easier)?

To an extent, yep. Though the only dupe bug I'm currently aware of is #205, which only happens in creative mode and seems to be caused by a client glitch.

PR #276, opened 2020-07-14
Fix entity duplication and add Velocity and PreviousVelocity to player

### LINKED ISSUE #275: Entity multiplication bug
This bug is quite severe, in that it duplicates or in some cases even triplicates (can happen when loading a vanilla world) every entity on the server, not in chunk (0,0), with every server restart. This makes the world unplayable and later even unloadable after a few server restarts. (I only tested it with item entities but the bug should be unrelated to the entity type)
This is caused mainly by two problems in the code base and while investigating I discovered a third problem which can lead to entity duplication:

1. The conversion from `Position` to `ChunkPosition` is wrong.
  https://github.com/feather-rs/feather/blob/026a7a6af6b6de4aeea9dac368e5bef22eda3d89/core/util/src/positions.rs#L177-L184
  The implementation as is uses integer division which floors the chunk position on positive indices and ceils it on negative ones. This actually causes entities in chunks with negative indices to be serialized into the wrong chunk.

2. New entities are created with a `PreviousPosition` of (0, 0, 0).
  https://github.com/feather-rs/feather/blob/026a7a6af6b6de4aeea9dac368e5bef22eda3d89/server/entity/src/lib.rs#L57-L64
  This causes an erroneous `ChunkCrossEvent` from chunk (0, 0) to the actual chunk to be processed immediatly after entity spawning, which in turn causes the same entity to be inserted a second time into `Game.chunk_entities`, which is queried by the serializer.
  https://github.com/feather-rs/feather/blob/026a7a6af6b6de4aeea9dac368e5bef22eda3d89/server/util/src/chunk_entities.rs#L7-L26

3. The third problem which could cause entity duplication (and actually caused the triplication in some cases in the vanilla world together with the first problem) is located in `feather-server-chunk::save::save_chunk_at`.
  https://github.com/feather-rs/feather/blob/026a7a6af6b6de4aeea9dac368e5bef22eda3d89/server/chunk/src/save.rs#L122-L124
  This early exit will prevent chunks which were loaded with entities in them from serializing without entities in case all of the entities have left the chunk before unloading without any blocks having been changed in the meantime.

I will begin working on a PR.
The first issue is trivial to fix. On the second issue I'll need some feedback in the PR (I plan on using an `Option` to wrap the position in `PreviousPosition` and also `PreviousVelocity` while I'm on it) and I am not quite sure how to remedy the third issue except for removing this optimization for now (which is not really desirable).

Schuwi:
This fixes #275 
I also added Velocity and PreviousVelocity components to player entities, I'm not sure whether they were intentionally left out but I couldn't find any evidence of them being added later so I took it as an oversight. Let me know whether I should revert that part.
Otherwise everything else should be explained in the issue.

### DIFF
--- a/server/entity/src/broadcasters/movement.rs
+++ b/server/entity/src/broadcasters/movement.rs
@@ -23,39 +23,42 @@ pub fn broadcast_movement(game: &mut Game, world: &mut World) {
         |(entity, (pos, prev_pos, id))| {
             let pos: Position = *pos;
 
-            if let Some(prev_pos) = prev_pos.0 {
-                if pos == prev_pos {
-                    return;
-                }
+            let prev_pos = match prev_pos.0 {
+                Some(prev_pos) => prev_pos,
+                None => return,
+            };
 
-                let entity_id = id.0;
+            if pos == prev_pos {
+                return;
+            }
 
-                let chunk = pos.chunk();
-                let players = game.chunk_holders.holders_for(chunk);
+            let entity_id = id.0;
 
-                for player in players.iter().filter(|player| **player != entity) {
-                    if let Some(network) = world.try_get::<Network>(*player) {
-                        let last_known_positions = world.get::<LastKnownPositions>(*player);
-                        let last_known_positions = last_known_positions.deref();
+            let chunk = pos.chunk();
+            let players = game.chunk_holders.holders_for(chunk);
 
-                        if let Some(mut last_known_pos) = last_known_positions.0.get_mut(&entity) {
-                            for packet in
-                                packets_for_movement_update(entity_id, *last_known_pos.value(), pos)
-                            {
-                                network.send_boxed(packet);
-                            }
+            for player in players.iter().filter(|player| **player != entity) {
+                if let Some(network) = world.try_get::<Network>(*player) {
+                    let last_known_positions = world.get::<LastKnownPositions>(*player);
+                    let last_known_positions = last_known_positions.deref();
 
-                            log::trace!("Updated position of {:?} on client {:?}", entity, player);
+                    if let Some(mut last_known_pos) = last_known_positions.0.get_mut(&entity) {
+                        for packet in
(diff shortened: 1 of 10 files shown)


server/entity/src/broadcasters/movement.rs
```diff
@@ -22,39 +22,40 @@ pub fn broadcast_movement(game: &mut Game, world: &mut World) {
         world.inner(),
         |(entity, (pos, prev_pos, id))| {
             let pos: Position = *pos;
-            let prev_pos: Position = prev_pos.0;
 
-            if pos == prev_pos {
-                return;
-            }
+            if let Some(prev_pos) = prev_pos.0 {
```
caelunshun, line 26: Just a minor style comment—this `if let` produces some more nested indentation. I'd prefer something like

```rust
let prev_pos = match prev_pos.0 {
     Some(prev_pos) => prev_pos,
    None => return,
}
```
to reduce nesting.

caelunshun:
It's unfortunate that we have to remove the chunk saving optimization, but I guess that's not a critical problem at the moment. I'll open an issue to track it.

Schuwi:
Okay, the newest commits should resolve #278 as well. It is kind of hard to test that though...

server/entity/src/lib.rs
```diff
@@ -38,17 +40,26 @@ pub fn previous_position_velocity_reset(world: &mut World) {
...
     <(Read<Velocity>, Write<PreviousVelocity>)>::query().par_for_each_mut(
         world.inner_mut(),
         |(vel, mut previous_vel)| {
-            previous_vel.0 = vel.0;
+            previous_vel.0.replace(vel.0);
         },
     );
 }
 
+#[fecs::event_handler]
```
caelunshun, line 54: This event handler will have to be added in the list in `server/src/event_handlers.rs` for it to be registered - right now, as far as I can tell, it will never be invoked.

PR #279, opened 2020-07-16
Change how commands handle successes, implement clear command

Molive-0:
Not sure if this should be two pull requests? I can split them if you want I guess.

Allows commands to return a Option<String> to provide a success statement, and implements the /clear command.

The ItemArgument class does not support item metadata as none of the rest of feather seems to either, that needs to be implemented. Apart from that this should work identically as to how the clear command works in vanilla 1.13.2 (based on switching between a singleplayer world and feather and comparing).


Molive-0:
Uh, this PL shouldn't change 122 files and add 54068 lines.
I'll look into that...?

Protonull:
probably a line endings thing?

Molive-0:
No, these are actually all new files! I've done something horrendously wrong here

Molive-0:
Okay, that should fix it. Not sure what happened there.

server/commands/src/arguments.rs
```diff
@@ -42,6 +43,45 @@ impl ArgumentKind<CommandCtx> for EntitySelector {
     }
 }
 
+impl EntitySelector {
+    /// Parses the returned entities for use in reporting success messages
+    /// Either the name of the entity for one entity, or how many were affected for many entities.
+    pub fn entities_to_string(&self, ctx: &CommandCtx, add_player: bool) -> String {
```
caelunshun, line 49: This could be moved to a `Display` impl so we can call `to_string()` or `format("{}", entities)` it.

server/commands/src/arguments.rs
```diff
@@ -300,3 +340,89 @@ impl AsRef<str> for TextArgument {
...
+#[derive(Debug, Error)]
+pub enum PositiveI32ParseError {
+    #[error("Invalid integer {0}")]
+    Invalid(String),
+    #[error("Integer must not be less than 0, found {0}")]
+    Negative(i32),
+}
+
+#[derive(Clone, Debug)]
+pub struct PositiveI32Argument(pub i32);
```
caelunshun, line 405: Out of curiosity - why do we need a `PositiveI32Argument` instead of just a `u32`? Is it for compatibility with vanilla, which always uses signed integers?

server/commands/src/impls.rs
```diff
@@ -28,7 +33,11 @@ pub fn tp_1(ctx: &mut CommandCtx, destination: EntitySelector) -> anyhow::Result
             teleport_entity_to_pos(&mut ctx.world, ctx.sender, pos);
         }
 
-        Ok(())
+        Ok(Some(format!(
+            "Teleported {0} to {1}",
+            ctx.world.try_get::<Name>(ctx.sender).unwrap().0.to_string(),
```
caelunshun, line 38: Instead of `try_get().unwrap()`, it's possible to call `get()` to unwrap the result automatically.

server/commands/src/arguments.rs
Molive-0: It is yes. Really what should be happening is that we call I32Argument but with a range argument which it checks before returning the value. Not sure how to do that

server/commands/src/arguments.rs
Molive-0: How can I implement Display/fmt and still have it pass the two parameters?
The "add_player" is probably not needed (makes the grammar the same as vanilla) but it needs the command context to work

server/commands/src/impls.rs
Molive-0: Think that's done :)

server/commands/src/arguments.rs
caelunshun: You're right, good point.

server/commands/src/arguments.rs
caelunshun: Hmm, that would require const generics, which are unstable. This will do fine, then :)

PR #280, opened 2020-07-18
Support non-x86 target compilation

Schuwi:
This PR implements a target-guard (`#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]`) in `feather-server-worldgen`'s noise.rs to only compile avx-related code on x86 targets.
~~For full non-x86 compilation support https://github.com/jackmott/rust-simd-noise/pull/31 has to be accepted to make the `simdnoise` dependency in `feather-server-worldgen` compile on non-x86 targets.~~
Edit: This PR was accepted way faster than expected, so feather should now successfully compile for non-x86 targets without any modifications to the source.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3016,9 +3016,9 @@ dependencies = [
 
 [[package]]
 name = "simdeez"
-version = "1.0.2"
+version = "1.0.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a0cc1807d0c7d2e869716572655b1abd32649b5600c047023f10c605bea8310f"
+checksum = "1893a4b8db694e23889b63229b6e1deb33846ee0552588bb6e45e53eff59667d"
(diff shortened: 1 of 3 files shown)


caelunshun:
Not sure why CI is failing—looks unrelated to this PR (missing crate `num_derive`?) I'll re-run it again and see if anything changes...

Schuwi:
Hmm running `cargo test` with `RUSTFLAGS="-C opt-level=0"` results in a stack overflow in tests::test_reproducability but at least it compiles...
I don't get this CI sometimes

Schuwi:
Actually running `RUSTFLAGS="-C opt-level=0" cargo test` with rustc v1.45 now passes all tests for me locally. Not really sure what else there would be different.

caelunshun:
The spurious compile failure seems to have disappeared, so I'll merge this now. Thanks!

PR #281, opened 2020-07-18
Implement SimplifiedBlockKind and refactoring

Schuwi:
`SimplifiedBlockKind` is a new enum parallel to BlockKind that aims to abstract away mostly unnecessary (e.g. only differing in texture) differences in vanilla BlockKind, with the primary aim of simplifying matches on the block kind.
For example something like
```
let block_id: BlockId = ...;
match block_id.kind() {
    BlockKind::WhiteBed
    | BlockKind::OrangeBed
    | BlockKind::MagentaBed
    [13 more of these just for the beds] => whatever(),
}
```
simplifies to
```
let block_id: BlockId = ...;
match block_id.simplified_kind() {
    BlockKind::Bed => whatever(),
}
```

This leads to far shorter, more readable, less error-prone match blocks where behaviour is dependant on the type of block, while still retaining the ability to get the specific block kind like when dropping an item for a destroyed block. Only a conversion from BlockKind to SimplifiedBlockKind is implemented because the other way around doesn't make any sense as it's losing information in the process. Right now the block types combined in SimplifiedBlockKind is kind of arbitrary, as I couldn't think of a good rule to go by.
In the process of this I also moved hardcoded properties of blocks dependent on the block kind to feather-blocks to have them centralized.

- Implement use of SimplifiedBlockKind throughout the codebase where appropriate
- Move some inline block properties to feather-core
- Fix typos in wall_blocks.rs and serialization_helper.rs

I am not quite sure whether all of this is the best approach to this problem of code duplication, so any feedback would be appreciated.


Inky-developer:
Doesn't the vanilla server group all those block kinds in json files? If so we could extract them in feather-data so that there is less hardcoded stuff and simpler version compatibility.

Schuwi:
Mmm yeah but tags aren't really the same thing. There can be multiple tags for the same block type. So you can't really compile a unique BlockKind => BlockTag mapping. The reason I am so hell-bent on keeping `match`es for this purpose everywhere is that it gives the compiler much more optimization options on how to implement that check on a case-by-case basis. Also I think having combinations of tens of `id.is_buttons() || id.is_beds() || id.is_banners()` doesn't look nearly as neat as a match, especially when the inevitable `|| id.kind() == BlockKind::Chest || id.kind() == BlockKind::Grass` joins the party as well. Sadly generating match arms from macros is impossible as well. Sooo I am just not sure what a good implementation with tags should look like.
Newer version compatibility should not be that big of a deal in this case, as SimplifiedBlockKind is generated from BlockKind and just uses regex to replace ".+_banner" with "banner", so new block types would be present, just not 'neatified'.

Inky-developer:
You are probably right. I just don't like that we would have an enum for eg all wool types and a set or list of the same data.

core/blocks/src/categories.rs
```diff
@@ -1,184 +1,75 @@
...
+    ChorusPlantLike,
+    MushroomLike,
+    SnowLike,
+    SugarCaneLike,
+    TripwireHookLike,
+    VineLike,
+}
 
 impl BlockId {
+    #[inline(always)]
```
caelunshun, line 34: What's the reason for the inlining here? I understand it might yield performances in some circumstances, but in general I'd prefer to leave it to the optimizer to decide on inlining. A compromise would be just `#[inline]`, which allows for cross-crate inlining but doesn't force the optimizer to do so. `#[inline(always)]` is generally discouraged in Rust, from what I can tell.

server/player/src/packet_handlers/placement.rs
```diff
@@ -215,12 +218,12 @@ fn handle_slab_placement(
     mut target_block_pos: BlockPosition,
     placement_face: PacketFace,
 ) -> bool {
-    if !block_to_place.is_slab() {
+    if !matches!(block_to_place.simplified_kind(), SimplifiedBlockKind::Slab) {
```
caelunshun, line 221: `block_to_place.simplified_kind() == SimplifiedBlockKind::Slab` would be more succinct here. The same applies in a few more lines in this diff.

Schuwi:
> You are probably right. I just don't like that we would have an enum for eg all wool types and a set or list of the same data.

I mean, there will probably be overlaps in the tags data and this enum but it won't be exactly the same and it's probably more work to generate this enum from the tags data than it is to generate it the way it's implemented in this PR. The partly duplicate data will be in the binary either way, it just changes where it's generated from.
Also, I don't really see where the tags data will see use except in sending the Tags packet to the client.

core/blocks/src/categories.rs
Schuwi: Hmm, yes, I wasn't that aware of the distinction between `#[inline]` and `#[inline(always)]`. `#[inline]` definitely makes more sense in this case then. Though I wouldn't remove the hint completely, because I suspect inlining these methods can make a lot of sense, especially when only forwarding to another function or matching on the return value (SupportType, PlacementType) in the caller.

server/player/src/packet_handlers/placement.rs
Schuwi: Not sure why I was so hellbent on that macro :D

Issue #157, opened 2019-10-06
Lava buckets place water


Ethan-Xd:
![image](https://user-images.githubusercontent.com/24442053/66267999-dbaf1880-e894-11e9-8ba1-ccd9eeca6d9b.png)

On the left is what a bucket of water places. On the right is what a bucket of lava places. Strangely, the right liquid is opaque, gives off light, and still makes lava sounds. I am not using any resource packs.

(Also, liquid physics don't seem to work correctly as reported in this release?)

caelunshun:
Yes, this issue has existed for a while. Probably the client expects us to send some sort of block update to tell them about the liquid, but Feather doesn't yet have support for buckets/liquids.

> (Also, liquid physics don't seem to work correctly as reported in this release?)
Sorry, that was worded somewhat poorly—entities now slow down when going through liquids, but liquid physics are not implemented yet.

amberkowalski:
Closed - Outdated

Issue #267, opened 2020-06-11
Discussion: how to ship vanilla data (recipes, loot tables) in Feather


caelunshun:
Loot tables and recipes are loaded from the vanilla data pack, which is extracted from the vanilla server JAR through `feather-data`. Currently, they are embedded in the server executable as JSON. This raises a number of legal issues, since we're directly packaging and redistributing Mojang assets. I see three potential solutions:

* Transcode the data from JSON to another format, e.g. CBOR. The assets will be distributed in a different format, making the legal violation slightly less blatant. However, I doubt this is legal either.
* Download the vanilla data pack at server startup. This will completely avoid packaging Mojang assets, but it's not very convenient for users to wait for the download to complete. Feather won't feel as "lightweight" when its first startup potentially takes a minute or two.
* Manually write the recipes and loot tables in our own format without consulting the vanilla assets. This will be extremely tedious, but it is the most ideal option in terms of results.

If anyone else has another idea, please share it here—none of the options above seem all that great to me :)

Schuwi:
I mean, ideally it will always be the same data in the end, so I guess for the legal aspect the thing that matters is where it comes from?
So at least for recipes community sources like https://minecraft.gamepedia.com/Crafting#Complete_recipe_list might be an idea but I doubt we'd find any community sourced loot tables that aren't just extracted from the Vanilla JAR.

Sooo... I had another idea, which is most certainly overkill and not really practical but I guess it's worth at least mentioning...
If one were to use a bot to connect to a vanilla server, run automated tests to source the info needed (again easy for recipes as they are just sent to the client I believe? but probably very hard for the loot table) and generate a handy file...

Pros:
- This should at least be perfectly legal (unless writing bots for Minecraft is generally forbidden or something, which could be the case as I've actually never read the EULA to be honest lol)
- No inconvenience for the user yay!!
- No need for error-prone laborious manual updating of recipes on updates (or in general)

Cons:
- Well... I mean no sane person would do this right?
- It's a *lot* of effort. though there are probably at least existing implementations for general purpose Minecraft bots that could be appropriated
- Instead of having to update a list of recipes you now have to update a whole client, yeah! (though it's probably mostly protocol which has to be updated anyway)

There are probably crazier things that have been done buuuut this would definitely require a good amount of desperation :D

Schuwi:
I just found out we get some of our stuff (like block states) by letting the vanilla server jar generate the (json?) files for us (https://wiki.vg/Data_Generators).
[...]

PR #282, opened 2020-07-18
Implement basics of Tags and HeldItemChangeClientbound packets

Schuwi:
This PR aims to implement the basics of the Tags packet - right now primarily used to make sure lava is rendered correctly in the client - and the HeldItemChangeClientbound packet - sent at server join to set the players selected hotbar slot to the one they had selected on their last login.
Tags are a new feature of 1.13 that specify common traits of block types/item types/fluid types and could be sourced from the vanilla Data Generators. For now I only hardcoded the fluid tags for lava and water to fix the vanilla client rendering lava with a water texture without a tags packet and instead intend to wait on a decision in #267 for a full-fledged implementation of tags.

### DIFF
--- a/core/network/src/packets.rs
+++ b/core/network/src/packets.rs
@@ -2423,29 +2423,8 @@ pub struct Tags {
 }
 
 impl Packet for Tags {
-    fn read_from(&mut self, buf: &mut Cursor<&[u8]>) -> anyhow::Result<()> {
-        for field in [
-            &mut self.block_tags,
-            &mut self.item_tags,
-            &mut self.fluid_tags,
-        ]
-        .iter_mut()
-        {
-            let length = buf.try_get_var_int()?;
-            for _tag in 0..length {
-                let identifier = buf.try_get_string()?;
-                let inner_length = buf.try_get_var_int()?;
-
-                let mut entries = Vec::with_capacity(inner_length as usize);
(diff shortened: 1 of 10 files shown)


core/network/src/packets.rs
```diff
@@ -2406,3 +2414,65 @@ pub struct EntityTeleport {
...
+            &mut self.fluid_tags,
+        ]
+        .iter_mut()
+        {
+            let length = buf.try_get_var_int()?;
+            for _tag in 0..length {
+                let identifier = buf.try_get_string()?;
+                let inner_length = buf.try_get_var_int()?;
+
+                let mut entries = Vec::with_capacity(inner_length as usize);
```
caelunshun, line 2439: Avoid calling `with_capacity` with untrusted inputs—it makes DoS attacks trivial :) A check here to ensure the length isn't too large would be good.

Schuwi: Thanks, good tip! Though I'll just remove the `read_from` implementation completely, it's never needed anyway and the new packet lib is coming up

Schuwi: Never needed as in this is actually a Server -> Client only packet

CDFN:
Apparently it closes #157 - can't reproduce after changes.

PR #284, opened 2020-07-20
Change from simdnoise to fastnoise_simd

pielover928:
Very similar implementations, saves 30 seconds on build time.


Schuwi:
fastnoise_simd is just "a mostly safe wrapper" for a C++ implementation instead of pure Rust though, right?

pielover928:
Yeah, libclang is required to build; I actually also made this PR earlier than I meant to. I was also going to look into bracket-noise, which is a complete rust port of fastnoise, but it doesn't seem to include SIMD support.

Schuwi:
Please also consider #280, (cross-)compilation support for non-x86 targets is a worthwhile goal to persue in my opinion. Not sure what impact a native C dependency would have on this

pielover928:
I have to boot up a VM to check, but SIMD should be okay compiling on ARM64/other architectures. A native solution would be better in every way so hopefully we get a native solution that's lighter at some point.

I'm going to close this for now (since it's not fully functional yet) but if I get it in a full-featured state I'll open it again.

PR #286, opened 2020-07-24
Fix updated clippy lint and update Cargo.lock

Schuwi:


Molive-0:
This seems like a good thing to merge

PR #288, opened 2020-07-28
Implement tool damage (nbt) when breaking blocks

Deliquescence:
This implements basic tool damage only when digging blocks--various right click forms of damage such as on bows or fishing poles are not implemented.

It also implements saving nbt tags on item entities and items in inventories, although the only one there so far is Damage.

Still todo is to actually break tools that have expended all durability. Would it be possible to include maximum durability in feather-definitions?


server/block/src/chest.rs
```diff
@@ -223,12 +219,8 @@ fn load_inventory(slots: &[InventorySlot]) -> Inventory {
     let inv = Inventory::chest();
 
     for slot in slots {
-        if let Some(item) = Item::from_identifier(&slot.item) {
-            if let Err(e) = inv.set_item_at(
-                Area::Chest,
-                slot.slot as usize,
-                ItemStack::new(item, slot.count as u8),
-            ) {
+        if let Some(_item_identifier) = Item::from_identifier(&slot.item) {
```
caelunshun, line 222: This could be changed to `if Item::from_identifier(&slot.item).is_some()` for clarity.

server/player/src/packet_handlers/inventory.rs
```diff
@@ -427,17 +424,16 @@ fn handle_double_click(
...
-        let item_type;
 
         // Get information about the currently picked item
-        if let Some(picked) = world.try_get_mut::<PickedItem>(player) {
-            stack_size = picked.0.ty.stack_size() as u8;
-            current_count = picked.0.amount;
-            item_type = picked.0.ty;
-        } else {
+        let picked = world.try_get_mut::<PickedItem>(player);
+        if picked.is_none() {
```
caelunshun, line 430: Rather than unwrapping the option after the `is_none()` check, this should be changed to a match:

```rust
let picked = match picked {
    Some(picked) => picked.0,
    None => return Ok(()),
}
```

Deliquescence:
Made those changes! In the future I'll take a look at manually doing the max damage definitions.

PR #290, opened 2020-08-02
Added the command generator and somewhat integrated the first result.

Miro-Andrin:
I have included the code i used to parse mojang's commands.json file. I put it at the top of the folder hierarchy in the folder command_generator. Not sure if that is where you want it, but that is were i put it. 

**Note:** I have also "integrated" the first result when  running the utility. The project compiles, but all the existing command handling code has been commented out and left where it was. That means **no commands are currently working.** 

I have run into two issues when integrating the generated code. First of all i have discovered some bugs in lieutenant that prevents me from integrating all the commands. That is why some functions in impls.rs and lib.rs are commented out. I am going to open a Issue. 
Another issue is the size of impls.rs, without adding any functionality it has reached 10_000 lines of code. In my editor, the highlighting/hints were unbearably slow. Maybe it should be split up, to make it more pleasant to work with.


Miro-Andrin:
The refactoring of #307 Into Minecraft 16.2 has moved were code generation happens. I am going to change this pull request, so that it better matches the new refactoring, once the changes are merged.

caelunshun:
> The refactoring of #307 Into Minecraft 16.2 has moved were code generation happens. I am going to change this pull request, so that it better matches the new refactoring, once the changes are merged.

Sounds good, thanks for your work on this!

PR #298, opened 2020-08-16
Implement /kick command

toiglak:
For #229 

If I missed something let me know! 😅


server/commands/src/impls.rs
```diff
@@ -287,6 +288,68 @@ pub fn me(ctx: &mut CommandCtx, action: TextArgument) -> anyhow::Result<()> {
...
+) -> anyhow::Result<Option<String>> {
+    for entity in &targets.entities {
+        if ctx.world.try_get::<Player>(*entity).is_none() {
+            return Err(KickError::NoEntities.into());
+        }
+    }
+    for entity in &targets.entities {
+        // Send kick packet and disconnect
+        let network = ctx.world.get::<Network>(*entity);
+        let kick_packet = DisconnectPlay {
```
caelunshun, line 330: This should use the function `game.disconnect()` instead, which properly removes the player from the world.

toiglak: Ok, for some reason I couldn't read backtrace... Took me long enough 👀

It turns out that there is a bug with function `handle_chat` (see below). First, we're dispatching a command and then logging it to the console. Logger takes as an argument `world.get::<Name>(player).0` but because that player might have been despawned by command it can't `get` player's name from `world` and panics.

That problem could be fixed by simply changing the order in which we call `commands.dispatch()` and `log::info!()`. 
I don't see a reason why that was ordered in this way in the first place, but if there is one let me know. 
Also, should I create a separate PR to fix it?

https://github.com/feather-rs/feather/blob/6fc2f2c0681168252fa664b9953eac98e599ed31/server/player/src/packet_handlers/chat.rs#L20-L26

caelunshun: Ah, that's the problem. Should be fine to make this change in this PR, since the functionality is related.

I think it's a bug that `game.disconnect()` doesn't properly send the kick packet; we should change it to send the packet with the disconnect reason in addition to logging to the console.

PR #299, opened 2020-08-16
Implement maximum durability and breaking used tools

Deliquescence:
This adds item durability to feather-definitions and uses that information to break items after the maximum durability has been expended (remove item from inventory and play sound).

In the future there may need to be particles spawned when some items break. In testing on vanilla server I only noticed this happening when flint and steel breaks and not when digging tools break, so I decided to wait on implementing the particles (since flint and steel isn't implemented yet).


server/entity/src/broadcasters/inventory.rs
```diff
@@ -126,6 +130,66 @@ fn is_equipment_update(held_item: usize, slot: SlotIndex) -> Result<Equipment, (
     }
 }
 
+/// System for damaging inventory items which should take damage.
+#[fecs::event_handler]
+pub fn on_damage_item(event: &ItemDamageEvent, game: &mut Game, world: &mut World) {
```
caelunshun, line 135: This function is pretty long—it would be good to split it up into separate parts.

Deliquescence: I pulled out a function for sending the sound effect, so it should be more reasonable now.

caelunshun:
Great, thanks!

PR #307, opened 2020-08-23
Upgrade to 1.16 + big refactor + initial plugin API

### LINKED ISSUE #220: Update MC implementation from 1.13.2 to 1.16.2
Here are the main changes in 1.15/1.14 which will affect us:
* Biomes are now 3D
* Chunks now store one section above and below `[0, 255]` for lighting purposes
* Lighting is now sent in a separate packet from `Chunk Data`, and it appears lighting updates are now sent to clients (previously, the server and client would each handle lighting on their own)

#307

### LINKED ISSUE #270: Implement smoother chunk streaming
The server is loading chunks too fast :D
Right now the server checks which new chunks to load the instant a player crosses a chunk border and then sends all the newly loaded chunks to the player as fast as it can.
When running a release build with a server view distance of 16 the Minecraft vanilla client suffers. Every time when crossing a chunk border the client hangs for a good ~100ms while receiving the new 32 chunks sent by the server. When setting the view distance to 32 the client actually seems to lose a lot of chunks (maybe some receive-buffer overflowing or something?), so a lot of chunks in view distance are just permanently unloaded for the client.
I actually wonder how the vanilla server handles this. Maybe it's just slow enough at loading chunks ^^?
I'm not sure how to remedy this. The straightforward solution would be to implement some kind of chunk streaming queue and only send x chunks per time interval t to the client but maybe there is a better solution?

caelunshun:
Upgrades the codebase to support ~~1.16.2~~ ~~1.16.3~~ 1.16.5 as the sole supported version.

Also, this PR includes a substantial refactor in terms of the crate structure. I'll elaborate more on this as the refactor progresses.

(Not even close to finished!)

Key changes (planned, not all implemented yet):
* [x] Implement `feather-protocol` as the successor to `feather-network`.
* [x] Use new `feather-generated` crate for protocol IDs and data (this crate replaces `feather-data` and `feather-definitions`)
* [x] Systems now return an `anyhow::Result` which can be an `Err` variant when an internal error occurs (e.g. an entity is missing a component, or a resource is not present.) Systems should no longer use `unwrap` or `panic` in most cases.
* [x] Move all gameplay-related systems into new `feather-common` crate.
* [x] `feather-server` now only contains systems needed to send and receive packets. All gameplay functionality is implemented in `feather-common`.
* ~~Add a new, experimental dynamic linking workflow that could help iteration times during development.~~

Resolves #220. 
Resolves #270.


crates/server/src/client.rs
```diff
@@ -116,11 +129,30 @@ impl Client {
             .set(self.teleport_id_counter.get() + 1);
     }
 
+    pub fn uupdate_own_chunk(&self, pos: ChunkPosition) {
```
ColdIce1605: miss spelling

ColdIce1605: damn it new commit fixes that :D

caelunshun:
I've just pushed an initial implementation of the barebones plugin API.

As this is becoming a colossal PR, I'm going to merge it in a few days, if nobody has objections. I think it's time we moved development back to the main branch :)

caelunshun:
Finally time to merge this!

Some features were lost in the rewrite, so I've moved the old code from before this PR into the `old` directory. Old functionality can be reimplemented with the aid of that old code.

PR #311, opened 2020-09-19
Implement seed command

CDFN:
Implements /seed command for #229 
Currently command's output isn't colored nor clickable as it is in vanilla.


caelunshun (approved):
Thanks, sorry about the delay!

PR #312, opened 2020-09-30
Add middle click handler

Koalab99:
If the player is in creative gamemode + has nothing in hand (PickedItem = None), then make a stack out of whatever he clicked.

Sometimes clicking doesn't work but it's because long click (of any
kind) are considered painting by the client and so by the server and it is not yet implemented.

Let me know if it needs a rework.

### DIFF
--- a/server/player/src/packet_handlers/inventory.rs
+++ b/server/player/src/packet_handlers/inventory.rs
@@ -596,14 +596,8 @@ fn handle_middle_click(
 
         // Pick the item in the slot
         let picked = accessor.item_at(packet.slot as usize)?;
-        let count = picked.map(|item| item.ty.stack_size());
-        if count.is_none() {
-            drop(accessor);
-            drop(window);
-            return Ok(());
-        }
-        let count = count.unwrap();
         if let Some(item) = picked {
+            let count = item.ty.stack_size();
             drop(accessor);
             drop(window);
             world
--- a/server/player/src/packet_handlers/inventory.rs
+++ b/server/player/src/packet_handlers/inventory.rs
@@ -205,7 +205,12 @@ impl<'a> TryFrom<&'a ClickWindow> for Mode {
 
                 Ok(Mode::NumberKey(value.button + 1))
             }
-            3 => Ok(Mode::MiddleClick),
+            3 => {
+                if value.button != 2 {
+                    return Err(ModeParseError::InvalidNumberKeyId(value.button));
+                }
+                Ok(Mode::MiddleClick)
+            }
             4 => {
                 if value.slot == -999 {
                     return Err(ModeParseError::Unhandled);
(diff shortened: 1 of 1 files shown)


server/player/src/packet_handlers/inventory.rs
```diff
@@ -206,7 +205,7 @@ impl<'a> TryFrom<&'a ClickWindow> for Mode {
 
                 Ok(Mode::NumberKey(value.button + 1))
             }
-            3 => Err(ModeParseError::Unhandled),
+            3 => Ok(Mode::MiddleClick),
```
Defman, line 208: Check if the button is 2.

server/player/src/packet_handlers/inventory.rs
```diff
@@ -574,6 +574,36 @@ fn handle_number_key(
...
+        if let Some(_) = world.try_get::<PickedItem>(player) {
+            // Player already has something in its hand.
+            return Ok(());
+        }
+        let window = world.get::<Window>(player);
+        let accessor = window.accessor(world)?;
+
+        // Pick the item in the slot
+        let picked = accessor.item_at(packet.slot as usize)?;
+        let count = picked.map(|item| item.ty.stack_size()).unwrap_or(0);
```
Defman, line 594: We cannot have a stack size of 0, should at least be one. Therefore if stack_size fails we should do nothing.

Defman (changes requested):
It looks good, just some potential minor issues.

server/player/src/packet_handlers/inventory.rs
Defman: In other words, just use unwrap. Since `stack_size()` can't return anything less than 1. You are welcome to update the return type of `Item::StackSize` to `NoneZeroU32`.

Defman: Alright, nvm, I'm wrong. Due to the `picked` being `None` this should be ok.

server/player/src/packet_handlers/inventory.rs
```diff
@@ -574,6 +579,42 @@ fn handle_number_key(
...
+        // Pick the item in the slot
+        let picked = accessor.item_at(packet.slot as usize)?;
+        let count = picked.map(|item| item.ty.stack_size());
+        if count.is_none() {
+            drop(accessor);
+            drop(window);
+            return Ok(());
+        }
+        let count = count.unwrap();
+        if let Some(item) = picked {
```
Defman, line 606: So what I was trying to suggest is to move the count into this if let.
```rust
if let Some(item) = picked {
  let count = item.ty.stack_size();
}
```

Defman (changes requested):

Defman (approved):

caelunshun (commented):
Thank you!

PR #315, opened 2020-10-04
removes java dependency for extracting server.jar

nobbele:


Defman:
The reason we used java for the extraction was due to also using the generator that is included with Minecraft. It seems that we have moved to manly use prismarinejs for our data needs, we might be able to remove the Minecraft jar completely.

1. I don't know if we are using the loot tables provided by Minecraft jar or prismarine js.
2. Prismarinejs does not include a translation table which might be used by the text API in the future.

caelunshun (approved):
Thanks! Yep, I forgot that JARs are just zip files...

PR #316, opened 2020-10-04
IP & UUID Ban support + Associated Commands

tcmal:
Implements /ban, /ban-ip, /pardon and /pardon-ip (#229). IP Bans are enforced before encryption is attempted, whereas UUID bans are enforced after authentication is performed.

When in online mode, `/pardon` uses the Mojang API to map the given name to a UUID.

Whilst ban expiry is supported, none of the implemented commands support adding expiring bans.

### DIFF
--- a/server/commands/src/impls.rs
+++ b/server/commands/src/impls.rs
@@ -10,13 +10,18 @@ use feather_core::text::{Text, TextComponentBuilder, TextValue};
 use feather_core::util::{Gamemode, Position};
 use feather_definitions::Item;
 use feather_server_types::{
-    ChatEvent, ChatPosition, GamemodeUpdateEvent, InventoryUpdateEvent, MessageReceiver, Name,
-    Player, ShutdownChannels, Teleported,
+    Ban, ChatEvent, ChatPosition, GamemodeUpdateEvent, InventoryUpdateEvent, MessageReceiver, Name,
+    Player, ShutdownChannels, Teleported, WrappedBanInfo,
 };
-use fecs::{Entity, ResourcesProvider, World};
+use feather_server_util::{name_to_uuid_offline, name_to_uuid_online};
+use fecs::{Entity, IntoQuery, Read, ResourcesProvider, World};
 use lieutenant::command;
 use smallvec::SmallVec;
+use std::net::{IpAddr, SocketAddr};
+use std::str::FromStr;
 use thiserror::Error;
+use tokio::runtime::Runtime;
+use uuid::Uuid;
 
 #[derive(Debug, Error)]
 pub enum TpError {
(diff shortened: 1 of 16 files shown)


server/commands/src/impls.rs
```diff
@@ -573,3 +578,242 @@ pub fn seed(ctx: &mut CommandCtx) -> anyhow::Result<()> {
...
+    Ok(None)
+}
+
+pub fn ban_players(
+    ctx: &mut CommandCtx,
+    targets: EntitySelector,
+    reason: String,
+    by_ip: bool,
+) -> anyhow::Result<Option<String>> {
+    if targets.entities.len() == 0 {
```
amberkowalski, line 688: All PRs in feather are run through clippy as part of the CI steps. This line fires a lint, unfortunately.
```
error: length comparison to zero
   --> server/commands/src/impls.rs:688:8
    |
688 |     if targets.entities.len() == 0 {
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: using `is_empty` is clearer and more explicit: `targets.entities.is_empty()`
    |
    = note: `-D clippy::len-zero` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#len_zero
```

For future reference, make sure you run `cargo clean` and `cargo clippy`. Sometimes clippy is funky without a `clean` first.

amberkowalski (commented):
I see a lot of `unwrap` throughout the PR. I'm not going to say that's bad, because I haven't looked into each unwrap. Obviously there are cases where a `None` or `Err` where it is reasonable to unwrap, however this is just a reminder to double check that the unwraps are reasonable. Otherwise everything else (other than the lint) looks fine, but I can't approve it because I just gave a cursory glance at it.

tcmal:
Most of the `unwrap`s come from using the RwLock. The only time this should panic is if another thread panics while it holds a lock, so hopefully this shouldn't be too much of an issue.

amberkowalski:
Rip, failed on formatting checks.

server/commands/src/impls.rs
```diff
@@ -573,3 +578,242 @@ pub fn seed(ctx: &mut CommandCtx) -> anyhow::Result<()> {
...
+    let online_mode = ctx.game.shared.config.server.online_mode;
+    let uuid = if online_mode {
+        Runtime::new()
+            .unwrap()
+            .block_on(name_to_uuid_online(&name.0))
+    } else {
+        Some(name_to_uuid_offline(&name.0))
+    };
+
+    if uuid.is_none() {
```
caelunshun, line 766: This could be more idiomatic with a match:

```rust
let uuid = match uuid {
    Some(uuid) => uuid,
    None => return Err(PardonError::NotPlayer.into()),
}
```

caelunshun (commented):
Thank you, this looks great! I'm sorry about the delay.

caelunshun:
Thanks!

PR #318, opened 2020-10-22
Fix lethal damage edge case

OmaMitStaubsauger:
I noticed two issues with the damage logic:
- an entity is only killed when the damage exceeds it's health, but not if it reaches exactly zero, while it should be considered dead at that point (I assume from the fact that the client shows the death screen at zero health, correct me if I'm wrong here) 
- when changing the threshold to less than one and reaching exactly zero health, the entities health is not updated to the new value before killing (which causes the client not to show a death screen )

EDIT: the second issue of course only occurred when fixing the first

### DIFF
--- a/server/types/src/game.rs
+++ b/server/types/src/game.rs
@@ -271,26 +271,15 @@ impl Game {
             return;
         }
 
-        let (should_kill, old_health, new_health) =
-            if let Some(mut health) = world.try_get_mut::<Health>(entity) {
-                let old_health = health.0;
-                let should_kill = match health.0.checked_sub(damage) {
-                    Some(0) => true,
-                    None => {
-                        // below 0
-                        health.0 = 0;
-                        true
-                    }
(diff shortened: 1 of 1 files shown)


OmaMitStaubsauger:
sorry, I noticed that I completely forgot to actually update the health component. I will fix it and write some tests :)

caelunshun (approved):
Thanks, looks good to me.

caelunshun:
Looks like there's a compile error from the borrow checker, at least on CI.

OmaMitStaubsauger:
Oh, I forgot to push the newest changes. I was also working on some test cases, but I'm struggling with the unit testing framework, so I can't include any for now.

I rebased on develop and checked `clippy` and `rustfmt` this time, but there are new changes unrelated to mine that broke the checks

caelunshun:
Thanks!

PR #317, opened 2020-10-20
Change stairs placement type to PlayerDirection

OmaMitStaubsauger:
Hey. I noticed that stairs were facing the wrong way when placing them, and this fixed it for me.

PR #319, opened 2020-10-23
satisfy clippy match_like_matches_macro lint

OmaMitStaubsauger:
Satisfies clippy by using the `matches!` macro where possible.
Also fixes a whitespace formatting error I introduced in #317 (sorry for that).


TheUserCreated:
The checks fail here due to github actions timing out during a tar, if run locally you will find that this does pass checks

caelunshun (commented):
Thanks.

PR #320, opened 2020-10-28
Drops falling block as item when landing on a non solid block

JonathanxD:
Currently, the falling block replaces the non solid block, such as torches and rails, with these changes, the falling block will be dropped as an item in the location it would land, just as it happens in Vanilla.

### DIFF
--- a/server/entity/src/object/falling_block.rs
+++ b/server/entity/src/object/falling_block.rs
@@ -1,6 +1,9 @@
 //! Implements falling block entities: sand, gravel, etc.
 
-use feather_core::blocks::{BlockId, SimplifiedBlockKind};
+use crate::drops::drop_item;
+use feather_core::item_block::BlockToItem;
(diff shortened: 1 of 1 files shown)


caelunshun (approved):
Thank you!

Schuwi:
There is a clippy warning in this commit that was merged:
```
error: redundant pattern matching, consider using `is_some()`
   --> server/entity/src/object/falling_block.rs:127:16
    |
127 |         if let Some(_) = not_solid_block_kind {
    |         -------^^^^^^^----------------------- help: try this: `if not_solid_block_kind.is_some()`
    |
    = note: `-D clippy::redundant-pattern-matching` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_pattern_matching
```

@caelunshun

PR #322, opened 2020-10-30
Time command

nobbele:
Implements /time command (#229)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1049,6 +1049,8 @@ dependencies = [
  "rand",
  "smallvec",
  "thiserror",
+ "tokio",
+ "uuid",
  "vek",
 ]
 
@@ -1110,6 +1112,7 @@ dependencies = [
  "derivative 2.1.1",
  "feather-core",
  "feather-server-types",
+ "feather-server-util",
  "fecs",
  "flume",
  "futures",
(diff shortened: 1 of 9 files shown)


durumu:
Hi, I'm not a contributor or anything, so I'm sorry if I'm out of my lane here. I was just looking through the issues and noticed a couple of things.

In Java edition, for whatever reason, time arguments are supposed to be parsed as `f32`s, not `u64`s: see [here](https://minecraft.gamepedia.com/Argument_types#minecraft:time).

Also, I think you're updating the world age, which should never get updated at all. In other words, `/time set 100.5d` shouldn't change the result of `/time query gametime`, only `/time query day` and `/time query daytime`. There should be two separate variables, one for the actual number of ticks the world has been through (world age) and one for the time as modified by console commands and sleeping. 

Also a couple of nitpicks. For one, I believe you forgot to update a comment on line 511 of `server/commands/src/arguments.rs`. For another, in `server/commands/src/impls.rs` I think you meant to put the includes in line 5 instead of line 3.

Other than that this looks good, and I hope to create something like this soon!

nobbele:
fixed everything mentioned by durumu

caelunshun:
Thanks!

PR #328, opened 2020-12-26
Fix writing negative i32 varints

PizzaCrust:
This fixes the issue when you write a negative integer and it infinitely loops and leaks. Additionally, I added more tests from wiki.vg


caelunshun:
Thanks! Looks like clippy added more lints in another Rust release, so the CI failure is unrelated.

PR #332, opened 2021-01-02
Implement sound effects for damage events

x4e:


core/network/src/packets.rs
```diff
@@ -1703,18 +1705,71 @@ impl Packet for PluginMessageClientbound {
...
     pub effect_pos_x: i32,
     pub effect_pos_y: i32,
     pub effect_pos_z: i32,
     pub volume: f32,
     pub pitch: f32,
 }
 
-#[derive(Clone, Copy)]
+impl Packet for NamedSoundEffect {
+	fn read_from(&mut self, buf: &mut Cursor<&[u8]>) -> anyhow::Result<()> {
```
aramperes, line 1720: Is reading necessary? It's an exclusively clientbound packet.

PR #339, opened 2021-01-14
Implement `/setworldspawn` command

Unrealrussian:
The command just updates the `LevelData` State in `ctx.game`. From which the next time `LevelData` is saved to disk will be permanent.
The `angle` parameter was not implemented, as it is only added in 1.16.

completes a point on #229


ColdIce1605:
@caelunshun close?

PR #343, opened 2021-02-08
text-api 2.0

Defman:
# Text-api 2.0
The text api now only support builder pattern like

## Todo
1. Test cases
2. Simplification
3. Cleanup and a few utility function that I don't know what to name.
4. Parsing of color codes

Also should we implement `From<Cow<'a, str>> for Text<'a>`?

### DIFF
--- a/crates/text/src/hover_event.rs
+++ b/crates/text/src/hover_event.rs
@@ -1,0 +1,14 @@
+use serde::{Serialize, Deserialize};
+use super::Text;
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", tag = "action", content = "value")]
+pub enum HoverEvent<'a> {
+    ShowText(Box<Text<'a>>),
+    ShowItem(ItemStack),
+    ShowEntity(LivingEntity),
+}
+
+pub type Uuid = ();
+pub type ItemStack = ();
+pub type LivingEntity = ();
--- a/crates/text/src/lib.rs
+++ b/crates/text/src/lib.rs
@@ -1,0 +1,365 @@
+mod click_event;
+mod color;
+mod content;
+mod hover_event;
+mod keybind;
+mod style;
+mod simplifier;
+
+pub use click_event::*;
+pub use color::*;
+pub use content::*;
+pub use hover_event::*;
+pub use keybind::*;
+pub use style::*;
+
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::borrow::Cow;
+
+/// A way to propperly serialize and deserialize extra components.
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(untagged)]
+enum Extra<'a> {
+    String(Cow<'a, str>),
+    Object(Text<'a>),
+    Array(Vec<Text<'a>>),
+}
+
+impl<'a> From<Text<'a>> for Extra<'a> {
+    fn from(text: Text<'a>) -> Self {
+        // Shoud be colapsed into a function on Text, just don't know what to name it.
+        // !has_atributes?
+        if text.color.is_none()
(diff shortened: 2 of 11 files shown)


crates/text/src/hover_event.rs
```diff
@@ -0,0 +1,14 @@
+use serde::{Serialize, Deserialize};
+use super::Text;
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", tag = "action", content = "value")]
+pub enum HoverEvent<'a> {
+    ShowText(Box<Text<'a>>),
+    ShowItem(ItemStack),
```
caelunshun, line 8: `ItemStack` is defined in `feather-generated` right now :)

crates/text/src/hover_event.rs
```diff
@@ -0,0 +1,14 @@
...
+    ShowEntity(LivingEntity),
+}
+
+pub type Uuid = ();
+pub type ItemStack = ();
+pub type LivingEntity = ();
```
caelunshun, line 14: We don't have a definition for `LivingEntity` AFAIK, but if we added one it would go into the region loading code.

crates/text/src/lib.rs
```diff
@@ -0,0 +1,365 @@
...
+enum Extra<'a> {
+    String(Cow<'a, str>),
+    Object(Text<'a>),
+    Array(Vec<Text<'a>>),
+}
+
+impl<'a> From<Text<'a>> for Extra<'a> {
+    fn from(text: Text<'a>) -> Self {
+        // Shoud be colapsed into a function on Text, just don't know what to name it.
+        // !has_atributes?
```
caelunshun, line 31: Maybe `is_plain`?

crates/text/src/lib.rs
```diff
@@ -0,0 +1,365 @@
...
+        }
+
+        Extra::Object(text)
+    }
+}
+
+struct ExtraVec;
+
+impl ExtraVec {
+    #[inline]
```
caelunshun, line 76: Get rid of these inlines please :)

crates/text/src/lib.rs
```diff
@@ -0,0 +1,365 @@
...
+    pub fn has_style(&self) -> bool {
+        self.bold.is_some()
+            && self.italic.is_some()
+            && self.obfuscated.is_some()
+            && self.strikethrough.is_some()
+            && self.underlined.is_some()
+    }
+
+    // is_bold? Bold can either be None, Some(false), and Some(true).
+    // Maybe a has_bold instead? self.bold.is_some()
```
caelunshun, line 251: I think each field of `Text` should be made private, and then we expose getter methods that abstract away this implementation detail.

crates/text/src/lib.rs
```diff
@@ -0,0 +1,365 @@
...
+            .unwrap(),
+        );
+    }
+
+    #[test]
+    fn json() {
+        let json = serde_json::to_string_pretty(&Text::of("hello world").red().bold()).unwrap();
+        println!("{}", json);
+    }
+}
```
caelunshun, line 365: Some test cases to consider:
* Just a JSON array - valid according to the wiki (empty root node with children from the array)
* Raw JSON primitives that aren't strings, like an integer

caelunshun (commented):
Looks great in general!

caelunshun:
@Defman we should probably transfer this PR to the `libcraft` repository :)

Defman:
I will move it, however I still have a few annoyances. Either we can keep our Text api close to raw json or we can make some modification and fix it in the serialization layer. One could remove `Option` from styles and color, but this would require context of where the text is used/from on the client. Since `None` indicate client default value. Removing extra and making text more flat could also be a consideration but that would make some other text operations harder.

In other words, I have a lot of open questions about how we should design and structure the TextApi.

PR #354, opened 2021-03-01
Add noop block placement

### LINKED ISSUE #353: Noop Block Placement
From: https://discord.com/channels/619316022800809995/805342116057579520/815963034462715975

Handle block placement packets from client and send a block change packet to the client to "undo" the block placement.

devmattrick:
Fixes #353. This add simple noop block placement by adding a packet handler and sending a packet to the player to undo that block placement.

Right now, libcraft doesn't have helper methods other than up() and down() on BlockPosition, so I am creating relative directions by hand and adding them to the position in the packet (the position in the PlayerBlockPlacement packet is the clicked block and not the actually placed block).

If feather-rs/libcraft#18 is merged then it can be done a bit more efficiently/ nicely.


Defman (approved):
Looks good, lets wait to merge it after the helper functions have been merged into libcraft. Did you test buckets, they may need to be handle slightly differently if I remember correctly.

devmattrick:
Updated to use the new libcraft methods. Buckets seem to work fine, the only issue is with doors at the moment, but this is blocked by #355.

caelunshun (approved):
Looks good to me. Thanks!

PR #361, opened 2021-03-02
Handle held item change packet

### LINKED ISSUE #355: Handle Held Item Change
Right now, feather has no way of knowing what item a player is holding because it doesn't handle the packets for it. This prevents feather from knowing what blocks are being placed, which tools are being used, etc.

feather should keep track of which slot the player's cursor is on, which can then be used to determine which item is being held based on the inventory state.

I'm not entirely sure where the best place to store this state in. Window seems like a possible candidate but it also seems too generic for this. It may make sense to make another component (HeldItem?) and attach that to the player.

devmattrick:
This adds basic handling to held item change packets. Fixes #355.

This can be used in the future to add "get item in hand" support but I think this needs to wait on a more complete player struct api?


crates/common/src/entities/player.rs
```diff
@@ -6,3 +6,7 @@ pub fn build_default(builder: &mut EntityBuilder) {
     super::build_default(builder);
     builder.add(Player).add(EntityKind::Player);
 }
+
+/// The hotbar slot a player's cursor is currently on
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct HotbarSlot(pub usize);
```
caelunshun, line 12: Could we encapsulate the `usize` to enforce that it's always in the range `0..9`? In other words, make the field private and add a setter method that returns a `Result`.

devmattrick: Done with d2a0eda

caelunshun (approved):
Thanks!

Issue #357, opened 2021-03-02
Entity physics


caelunshun:
Implement entity physics: gravity, collisions with blocks.

The basic algorithm for Minecraft entity physics is, each tick, to:
1. Add velocity to position.
2. For each axis (X/Y/Z) individually, check for collisions between the entity's bounding box and blocks. If there are any, clamp the entity's position to the edge of the block. There are some useful details on collision detection [here](https://www.reddit.com/r/gamedev/comments/qnmqk/good_collision_detection_sliding_on_walls/), there I'm sure there are better sources.
3. Apply gravity to velocity.
4. Multiply velocity by a drag factor.

In Minecraft physics, different entities behave differently with different drag/gravity forces. This parameters are listed [here](https://minecraft.gamepedia.com/Entity#Motion_of_entities). To implement this in Feather, we would use a `Physics` component containing a field for each parameter: `gravity_acceleration`, `drag_factor`, `bounding_box`, and `drag_before_acceleration: bool`.

lolgeny:
I'll try doing this as a first contribution

OxKitsune:
I think someone was working on this already, @Stoozy. I'm not sure how far they are but maybe you can collaborate?

lolgeny:
Oh ok, I'll do something else, I didn't know that

Stoozy:
Yea

PR #362, opened 2021-03-02
Add entity interaction event.

amberkowalski:
Adds an entity interaction event that gets fired when the server gets an entity interact packet. 

This is my first PR on the refactor so let me know if I put code in the wrong places or did anything wrong please.

TODO(?)
- [ ] The server is *supposed* to verify that the interaction is "valid" before it actually does anything. 
- [x] Sprinkle in some docs
- [x] Test this properly when entities can be spawned ~~(I can't figure out how to, there might already be a method of doing so)~~

### DIFF
--- a/crates/common/Cargo.toml
+++ b/crates/common/Cargo.toml
@@ -19,3 +19,4 @@ quill-common = { git = "https://github.com/feather-rs/quill", rev = "d4da5b6" }
 smartstring = "0.2"
 utils = { path = "../utils", package = "feather-utils" }
 uuid = { version = "0.8", features = [ "v4" ] }
+libcraft-core = { git = "https://github.com/amberkowalski/libcraft.git", branch = "amber-entity-interact" }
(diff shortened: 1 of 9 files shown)


amberkowalski:
Just did a little testing, it appears to work.

caelunshun (approved):
LGTM, thanks!

As for validation—we should probably wait for a physics implementation (#357) so we have the necessary bounding box methods.

amberkowalski:
mmmmm that's a mess. Anyway I got the kick working, and it returns a system error as well.

crates/common/Cargo.toml
```diff
@@ -19,3 +19,4 @@ quill-common = { git = "https://github.com/feather-rs/quill", rev = "d4da5b6" }
 smartstring = "0.2"
 utils = { path = "../utils", package = "feather-utils" }
 uuid = { version = "0.8", features = [ "v4" ] }
+libcraft-core = { git = "https://github.com/amberkowalski/libcraft.git", branch = "amber-entity-interact" }
```
amberkowalski, line 22: This needs to be updated to use the correct repo.

crates/server/Cargo.toml
```diff
@@ -47,6 +47,7 @@ ureq = { version = "2", features = [ "json" ] }
 utils = { path = "../utils", package = "feather-utils" }
 uuid = "0.8"
 vec-arena = "1"
+libcraft-core = { git = "https://github.com/amberkowalski/libcraft.git", branch = "amber-entity-interact" }
```
amberkowalski, line 50: This also needs to be updated to use the correct repo.

caelunshun (approved):
Thanks!

Issue #344, opened 2021-02-09
Plugin API Roadmap (Quill)


caelunshun:
#307 introduces a preliminary draft of Feather's plugin API, `quill`. This is tracking issue for features needed in the plugin API to achieve the goals laid out in [our mid-term roadmap](https://github.com/feather-rs/feather/issues/341).

Foundational features:
* [x] Plugin state struct
* [x] Access to `Copy` components from the host
* [ ] Mutating `Copy` components from the host
* [ ] Defining, accessing, and mutating custom components
* [ ] Creating and removing entities
* [ ] Access to opaque, non-`Copy` components from the host
  * [ ] `Name`
  * [ ] `Inventory`
  * [ ] `Window`

Features needed for a hub plugin implemented with Quill:
* [ ] Support all vanilla entities in Feather: armor stands, mobs
* [ ] Particles
* [ ] Others?

Features needed for minigames implemented with Quill:
* [ ] Block access, ideally with bulk operation support ("fill this chunk section with the given block")
* [ ] Custom `WorldSource`s defined in plugins for custom world loading/saving logic
* [ ] Others?

Miscallaneous:
* [ ] Plugin assets - stored in the plugin's `tar` file?

OxKitsune:
@caelunshun I'd like to tackle the particle system if that's possible :)

caelunshun:
> @caelunshun I'd like to tackle the particle system if that's possible :)

Sure thing, thanks!

We'll probably want a `Particle` type in [libcraft](https://github.com/feather-rs/libcraft) that defines all possible particles, plus their associated metadata.

The steps for implementing this are:
* Add support for spawning particles in [`quill`](https://github.com/feather-rs/quill). That means a `particle_spawn` method in the `quill-sys` crate, plus a method on `quill::Game` that invokes the raw `quill_sys::particle_spawn`.
* Implement the `particle_spawn` host call in Feather. This involves adding the function to `crates/plugin-host/src/host_calls.rs`.
* Add whatever is needed to send particles to clients.

Let me know if you have any questions, either here or on Discord.

OxKitsune:
@caelunshun I have a question on the design. Are you planning on spawning ecs-entities into the ecs system directly through quill, or should it be done like the entity builder is. Where we just pass a struct containing all the required data back from the client (the plugin/quill) to the host (feather) and create the ecs-entity there.

caelunshun:
[...]

PR #364, opened 2021-03-02
Implement particle host calls

OxKitsune:
This is the last PR for the particle implementation!
This PR implements the host calls required by quill (feather-rs/quill#4).
And should also wrap up the particle bullet point in #344 

It registers a System that looks for ecs-entities with the `Particle` and `Position` component, it then proceeds to send a particle packet to nearby players.


caelunshun (approved):
Thanks!

PR #367, opened 2021-03-03
Block interaction and placement events

amberkowalski:
This adds the `BlockPlacementEvent` which gets fired when a player sends a block placement packet to the server. I don't like how it's handled right now, since someone wanting to hook into this event would be confused as to why its a placement event and not an interact event. Ideally we could determine whether a block is being interacted with, or we should place a block. This PR is not ready to merge until we discuss that.

### DIFF
--- a/crates/common/src/events/block_placement.rs
+++ b/crates/common/src/events/block_placement.rs
@@ -1,0 +1,9 @@
+use libcraft_core::{BlockPosition, InteractBlockFace, InteractHand, Vec3f};
+
+pub struct BlockPlacementEvent {
+    pub hand: InteractHand,
+    pub location: BlockPosition,
+    pub face: InteractBlockFace,
+    pub cursor_position: Vec3f,
+    pub inside_block: bool,
+}
--- a/crates/server/src/packet_handlers/interaction.rs
+++ b/crates/server/src/packet_handlers/interaction.rs
@@ -1,8 +1,8 @@
 use crate::{ClientId, NetworkId, Server};
-use common::events::InteractEntityEvent;
+use common::events::{BlockPlacementEvent, InteractEntityEvent};
 use common::Game;
-use ecs::{Entity, EntityRef, SysResult};
-use libcraft_core::{InteractHand, InteractionType, Vec3f};
+use ecs::{Entity, SysResult};
+use libcraft_core::{InteractBlockFace, InteractHand, InteractionType, Vec3f};
 use protocol::packets::client::{
     BlockFace, InteractEntity, InteractEntityKind, PlayerBlockPlacement, PlayerDigging,
     PlayerDiggingStatus,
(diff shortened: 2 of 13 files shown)


crates/common/src/events/block_placement.rs
```diff
@@ -0,0 +1,9 @@
+use libcraft_core::{BlockPosition, InteractBlockFace, InteractHand, Vec3f};
+
+pub struct BlockPlacementEvent {
+    pub hand: InteractHand,
+    pub location: BlockPosition,
+    pub face: InteractBlockFace,
+    pub cursor_position: Vec3f,
+    pub inside_block: bool,
```
Defman, line 8: Is this if the player tried to place a block while he was inside a block?
Documentation.

crates/common/src/events/block_placement.rs
```diff
@@ -0,0 +1,9 @@
+use libcraft_core::{BlockPosition, InteractBlockFace, InteractHand, Vec3f};
+
+pub struct BlockPlacementEvent {
+    pub hand: InteractHand,
```
Defman, line 4: Could this not just be a generic Hand enum?

crates/common/src/events/block_placement.rs
```diff
@@ -0,0 +1,9 @@
...
+    pub location: BlockPosition,
+    pub face: InteractBlockFace,
```
Defman, line 6: Just generic block face?

crates/server/src/client.rs
```diff
@@ -469,6 +472,13 @@ impl Client {
     fn send_packet(&self, packet: impl Into<ServerPlayPacket>) {
         let _ = self.packets_to_send.try_send(packet.into());
     }
+
+    pub fn disconnect(&self, reason: &str) {
+        self.disconnected.set(true);
+        self.send_packet(Disconnect {
```
Defman, line 478: Could client not just intercept all disconnect packets?

amberkowalski: This isn't really part of this PR, and setting disconnected kills the tcp stream.

crates/server/src/systems.rs
```diff
@@ -78,5 +81,13 @@ fn tick_clients(_game: &mut Game, server: &mut Server) -> SysResult {
     for client in server.clients.iter() {
         client.tick();
     }
+
+    static COUNTER: std::sync::atomic::AtomicUsize = AtomicUsize::new(0);
+    if COUNTER.fetch_add(1, std::sync::atomic::Ordering::SeqCst) >= 200 {
+        for client in server.clients.iter() {
```
caelunshun, line 87: Looks like this commit got in here again :)

amberkowalski: I think its a compelling feature!

crates/server/src/packet_handlers/interaction.rs
```diff
@@ -1,35 +1,75 @@
...
+        BlockFace::Bottom => LibcraftBlockFace::Bottom,
+    };
+
+    let cursor_position = Vec3f::new(
+        packet.cursor_position_x,
+        packet.cursor_position_y,
+        packet.cursor_position_z,
+    );
+
+    let block_kind = { game.block(packet.position).unwrap().kind() };
```
caelunshun, line 42: This could panic if the block is unloaded, so we should have a check here.

crates/server/src/packet_handlers/interaction.rs
```diff
@@ -1,35 +1,75 @@
...
-        BlockFace::Bottom => packet.position.down(),
-        BlockFace::Top => packet.position.up(),
-        BlockFace::North => packet.position.north(),
-        BlockFace::South => packet.position.south(),
-        BlockFace::West => packet.position.west(),
-        BlockFace::East => packet.position.east(),
+    let hand = match packet.hand {
+        0 => Hand::Main,
+        1 => Hand::Offhand,
+        _ => unreachable!(),
```
caelunshun, line 24: Is this really unreachable? Couldn't the client send an invalid value?

crates/server/src/packet_handlers/interaction.rs
amberkowalski: I'm just gonna kick the player if they do that then.

crates/server/src/packet_handlers/interaction.rs
amberkowalski: Also going to kick the player here

caelunshun (approved):
Thanks!

PR #370, opened 2021-03-03
Moves interaction events to quill

amberkowalski:
Requires the quill and libcraft PR for tests to pass. There might also be issues with locked revisions on the git deps but I promise this builds on my machine and will build once we set up the monorepo.


amberkowalski:
Can you re-run CI?

Issue #12, opened 2019-07-04
Implement timeouts


caelunshun:
Currently, status pings which end off at Response rather than having the client send a ping packet remain permanently registered. A timeout should be implemented to deregister these network handles after some period of inactivity.

caelunshun:
Also, clients need to be kicked when keepalives aren't received for some time

caelunshun:
I'm going to delay this until the networking code is rewritten using Tokio, which will make this significantly easier to implement.

PR #373, opened 2021-03-04
ItemStack refactor (Continuation)

PauMAVA:
Continuation of the original `libcraft` repo PR [#12](https://github.com/feather-rs/libcraft/pull/12).

> This draft PR is related to [#6](https://github.com/feather-rs/libcraft/issues/6). We can discuss the implementation of the new ItemStack type here.


PauMAVA:
Still very WIP. Also, some metadata depends on `Dimensions`, `Entities`, and `Effects` some of them aren't implemented yet. Also, every metadata type must have its respective `getters`, `setters`, `defaults` etc. I don't think I will use `proc_macro` or `macro` for this as every meta-type has its custom limitations that differ from other types. Also will probably create a module for item meta and separate all meta types into different files so that we don't end up with a 2000+ lines file.

PR #374, opened 2021-03-04
Implement plugin messaging

### LINKED ISSUE #365: Send plugin message packets in plugins
Allow plugins to use the [plugin messaging system](https://dinnerbone.com/blog/2012/01/13/minecraft-plugin-channels-messaging/). Critically, plugin messaging would allow plugins to transfer players to different proxy servers using the [BungeeCord plugin messaging API](https://www.spigotmc.org/wiki/bukkit-bungee-plugin-messaging-channel/).

Steps to implement this:
1. Add a new host call called `plugin_message_send` to [`quill-sys`](https://github.com/feather-rs/quill).
2. Wrap the host call with a new method `Game::send_plugin_message(channel: &str, data: &[u8])` in `quill`.
3. Implement the host call in `feather-plugin-host` by triggering a new `PluginMessageEvent` (`game.ecs.insert_event`)
4. Add an event handler to `feather-server` that sends the `PluginMessage` packet to the client.
5. Test.

(For new contributors: I recommend reading [the architecture docs](https://github.com/feather-rs/feather/blob/main/docs/architecture.md) to get started.)

dlee13:
Closes #365 
Related to #344


feather/plugin-host/src/host_calls/plugin_message.rs
```diff
@@ -0,0 +1,24 @@
...
+#[host_function]
+pub fn plugin_message_send(
+    cx: &PluginContext,
+    channel_ptr: PluginPtr<u8>,
+    channel_len: u32,
+    data_ptr: PluginPtr<u8>,
+    data_len: u32,
+) -> anyhow::Result<()> {
+    let channel = cx.read_string(channel_ptr, channel_len)?;
+    let data = cx.read_string(data_ptr, data_len)?.into();
```
caelunshun, line 15: This will fail when the data isn't valid UTF-8, so we need to get a `Vec<u8>` directly rather than going through `String`.

dlee13: Should I add a `read_bytes` method?

caelunshun: Yes, that would make sense.

feather/plugin-host/src/host_calls/plugin_message.rs
```diff
@@ -0,0 +1,24 @@
...
+    data_len: u32,
+) -> anyhow::Result<()> {
+    let channel = cx.read_string(channel_ptr, channel_len)?;
+    let data = cx.read_bytes(data_ptr, data_len)?;
+
+    let event = PluginMessageEvent {
+        channel,
+        data,
+    };
+    cx.game_mut().ecs.insert_event(event);
```
caelunshun, line 21: Plugin messages should be associated with a single entity/player for the query on `ClientId` to work. We should add a new entity parameter (`u64`) to the host call that indicates the player to send the plugin message to. Then use `insert_entity_event(Entity::from_bits(entity_u64), event)`.

dlee13:
Would it be appropriate to make it a method on `Entity`?

Schuwi:
I don't feel like that would be appropriate as only entities representing a player are able to receive plugin messages.

OxKitsune:
We just had a discussion about this in the discord server, and @Defman mentioned that it might be a good idea to have a trait that can specify required components. Which can then be implemented for JUST players that require features such as chat and plugin messaging.

dlee13:
It is similar to `Entity::send_message` which also does nothing if the entity is not a player.

Schuwi:
Yeah, I would argue that `send_message` is misplaced as well, but @Defman proposed a good solution in Discord imo. Given the case it can be implemented without too much compile-time overhead. But that would be a separate PR either way.

caelunshun:
The ideal API needs to be as simple as possible, so the API that looks like `(entity, player).send_message()` isn't great. Exposing it as a method on the `Player` component is fine by me, though.

dlee13:
The `Player` component is in quill-common which both api and sys depend on, so putting it there does not seem to be an option.

caelunshun:
> The `Player` component is in quill-common which both api and sys depend on, so putting it there does not seem to be an option.

All right. I think we can merge this PR as is, then. Once we implement a better way to expose entity-specific functions like this, we can move the function to a more suitable place.

caelunshun (approved):
Thanks!

PR #376, opened 2021-03-04
DynamicBundle

### LINKED ISSUE #375: DynamicBundler, quill spawn entity shorthand
A spawn function which is similar to `hecs::World::spawn`

```rust
trait DynamicBundle {
  fn add_to_builder(self, builder: &mut EntityBuilder);
}

// Should be a macro
impl<T1> DynamicBundle for (T1,) where T1: Component {}
impl<T1, T2> DynamicBundle for (T1, T2) where T1: Component, T2: Component {}

struct ECS {
  fn spawn(bundle: impl DynamicBundle) -> Entity;
  fn insert_event_with_components(event: impl Component, bundle: impl DynamicBundle);
}
```

Defman:
Might need some documentation, I'm not familiar with the that @caelunshun is using.
Resolves #375


Defman:
The `Game::spawn` could also be optimized to not use entity builder and reduce the number of host calls required.

quill/api/src/game.rs
```diff
@@ -97,6 +98,15 @@ impl Game {
         EntityBuilder::new(id)
     }
 
+    /// Spawns a new [`Entity`] with the supllied components.
+    ///
+    /// The builder is not initialized with any default components.
+    pub fn spawn(&self, components: impl DynamicBundle) -> Entity {
```
caelunshun, line 104: Could we rename this to `spawn_entity_bundle` or similar? We already have `spawn_entity`, which is higher level and probably more frequently used.

Defman: What if I make `spawn` more generic, and make all components `DynamicBundle` and make the `tuple_impl` over `DynamicBundle`. This way `spawn_entity` would become `spawn(DefaultBundle)` or with more data attached `spawn((DefaultBundle::new(...), Health::default())`

caelunshun: `DefaultBundle` would have to be special cased in the server if we do this. There are a number of internal components inside the server that aren't exposed to plugins, so plugins aren't directly able to add all the default components for an entity.

Defman: I was not aware of this, the DefaultBundle have full access to the EntityBuilder so are there no way we can handle it in Quill?

caelunshun: EntityBuilder in Quill can only access components available to plugins, but many components aren't exposed to plugins, because they're implementation details. (`SpawnPacket`, `NetworkId`, a few others). Also, we might want to add new internal components in the future, and hardcoding the default component sets in plugins would make this non-backwards-compatible. In short, the server needs to be aware of the type of entity we're creating so it can initialize default components.

Defman:
I would like to have a single `spawn` function on game, that can take `PiglinBundle`, `Component`, and or any combination of components and bundles. Without specialisation there is no way to make a blank `ComponentBundle` implementation for `Component`s, I suggest that `Component: ComponentBundel` such that a `Component` is also a bundle, but it only adds it self to the entity.

quill/common/src/bundle.rs
```diff
@@ -0,0 +1,45 @@
+use crate::Component;
+
+pub trait EntityBuilder {
```
Defman, line 3: Should maybe be named something like `BundleSpawner` or `BundleAdder` or maybe straight up make the entity builder a trait.

Defman:
If this is an acceptable solution, the next step for me would be to create bundles for all `EntityInit` variants. Maybe update the `EntityBuilder` to only require one host call, instead of a host call to initialize the builder and for each added component.

Defman:
@caelunshun where did we land on this. Right now I have moved bundles into common such that I can make a blanket `ComponentBundle` implementation for `Component`. Should the function still be renamed to `spawn_bundle` or should we move to exclusively use bundles?

PR #377, opened 2021-03-04
Fixed server crash when joining with 1.8.

FussballAndy:
The server crashed at executing send_loaded_chunks feather/server/src/systems/view.rs if someone in 1.8. joined.

### DIFF
--- a/feather/server/src/initial_handler.rs
+++ b/feather/server/src/initial_handler.rs
@@ -58,7 +58,7 @@ pub async fn handle(worker: &mut Worker) -> anyhow::Result<InitialHandling> {
 
     let ClientHandshakePacket::Handshake(handshake) = handshake;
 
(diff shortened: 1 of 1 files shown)


feather/server/src/initial_handler.rs
```diff
@@ -58,6 +58,17 @@ pub async fn handle(worker: &mut Worker) -> anyhow::Result<InitialHandling> {
 
     let ClientHandshakePacket::Handshake(handshake) = handshake;
 
+    if handshake.protocol_version < PROTOCOL_VERSION {
```
caelunshun, line 61: We should do this only if the next state is `Login`, because otherwise we disconnect status pings on older versions, which should still go through.

caelunshun (approved):
Thanks!

PR #379, opened 2021-03-05
Move text to libcraft

OxKitsune:
This moves our text implementation to libcraft.

Clippy fails on something I didn't touch, so I'm not sure what's up with that, namely libcraft-items.


caelunshun (approved):
Thanks!

PR #387, opened 2021-03-06
Added dimensions enum.

### LINKED ISSUE #386: Adding Dimension enum on Libcraft
We should add an enum containing the dimensions in the core crate of libcraft.

I will try to implement this.

Masplus:
Implemented enum for dimensions

Created the functions to get the `dim_id` and the `namespaced_id`. 

Implemented TryFrom and Into traits, to serialize and deserialize using the `namespaced_id`.

Closes #386

### DIFF
--- a/libcraft/core/src/dimension.rs
+++ b/libcraft/core/src/dimension.rs
@@ -63,18 +63,6 @@ impl From<Dimension> for &'static str {
     }
 }
 
-impl TryFrom<i32> for Dimension {
-    type Error = &'static str;
-
-    fn try_from(value: i32) -> Result<Self, Self::Error> {
-        if let Some(val) = Self::from_id(value) {
-            Ok(val)
-        } else {
-            Err("Unknown dimension dim_id.")
-        }
-    }
-}
-
 impl From<Dimension> for i32 {
     fn from(value: Dimension) -> Self {
         value.id()
--- a/libcraft/core/src/dimension.rs
+++ b/libcraft/core/src/dimension.rs
@@ -35,8 +35,8 @@ impl Dimension {
         }
     }
 
-    pub fn from_namespaced_id(id: &String) -> Option<Self> {
-        match id.as_str() {
+    pub fn from_namespaced_id(id: &str) -> Option<Self> {
+        match id {
             "minecraft:overworld" => Some(Self::Overworld),
             "minecraft:the_nether" => Some(Self::TheNether),
             "minecraft:the_end" => Some(Self::TheEnd),
@@ -46,23 +46,37 @@ impl Dimension {
 }
 
 impl TryFrom<String> for Dimension {
-    type Error = String;
+    type Error = &'static str;
 
     fn try_from(namespaced_value: String) -> Result<Self, Self::Error> {
-        if let Some(val) = Self::from_namespaced_id(&namespaced_value) {
+        if let Some(val) = Self::from_namespaced_id(namespaced_value.as_str()) {
             Ok(val)
         } else {
-            Err(format!(
-                "There is no dimension with namespaced_id {}",
-                namespaced_value
-            )
-            .to_string())
+            Err("Unknown dimension namespaced_id.")
         }
     }
 }
 
-impl Into<&'static str> for Dimension {
-    fn into(self) -> &'static str {
-        self.namespaced_id()
+impl From<Dimension> for &'static str {
+    fn from(value: Dimension) -> Self {
+        value.namespaced_id()
+    }
+}
+
+impl TryFrom<i32> for Dimension {
+    type Error = &'static str;
+
+    fn try_from(value: i32) -> Result<Self, Self::Error> {
+        if let Some(val) = Self::from_dim_id(value) {
+            Ok(val)
+        } else {
+            Err("Unknown dimension dim_id.")
+        }
+    }
+}
+
+impl From<Dimension> for i32 {
+    fn from(value: Dimension) -> Self {
+        value.dim_id()
     }
 }
--- a/libcraft/core/src/dimension.rs
+++ b/libcraft/core/src/dimension.rs
@@ -10,7 +10,7 @@ pub enum Dimension {
 }
 
 impl Dimension {
-    pub fn dim_id(&self) -> i32 {
+    pub fn id(&self) -> i32 {
         match self {
             Self::Overworld => 0,
             Self::TheNether => -1,
@@ -18,7 +18,7 @@ impl Dimension {
         }
     }
 
-    pub fn from_dim_id(id: i32) -> Option<Self> {
+    pub fn from_id(id: i32) -> Option<Self> {
         match id {
             0 => Some(Self::Overworld),
             -1 => Some(Self::TheNether),
@@ -67,7 +67,7 @@ impl TryFrom<i32> for Dimension {
     type Error = &'static str;
 
     fn try_from(value: i32) -> Result<Self, Self::Error> {
-        if let Some(val) = Self::from_dim_id(value) {
+        if let Some(val) = Self::from_id(value) {
             Ok(val)
         } else {
             Err("Unknown dimension dim_id.")
@@ -77,6 +77,6 @@ impl TryFrom<i32> for Dimension {
 
 impl From<Dimension> for i32 {
     fn from(value: Dimension) -> Self {
-        value.dim_id()
+        value.id()
     }
 }
--- a/libcraft/core/src/dimension.rs
+++ b/libcraft/core/src/dimension.rs
@@ -1,0 +1,68 @@
+use serde::{Deserialize, Serialize};
+use std::convert::TryFrom;
+
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(try_from = "String", into = "&'static str")]
+pub enum Dimension {
+    Overworld,
+    TheNether,
+    TheEnd,
+}
+
+impl Dimension {
+    pub fn dim_id(&self) -> i32 {
+        match self {
+            Self::Overworld => 0,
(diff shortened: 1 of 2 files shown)


libcraft/core/src/dimension.rs
```diff
@@ -0,0 +1,68 @@
...
+        match id.as_str() {
+            "minecraft:overworld" => Some(Self::Overworld),
+            "minecraft:the_nether" => Some(Self::TheNether),
+            "minecraft:the_end" => Some(Self::TheEnd),
+            _ => None,
+        }
+    }
+}
+
+impl TryFrom<String> for Dimension {
```
Defman, line 48: Should be `FromStr`

libcraft/core/src/dimension.rs
```diff
@@ -0,0 +1,68 @@
...
+            Err(format!(
+                "There is no dimension with namespaced_id {}",
+                namespaced_value
+            )
+            .to_string())
+        }
+    }
+}
+
+impl Into<&'static str> for Dimension {
```
Defman, line 64: Prefer `From`

Defman (changes requested):

OxKitsune:
Not sure how I feel about the name "namespaced_id". We call it "name" everywhere else, but I think namespaced id makes more sense.

Masplus:
I'll fix it tomorrow 😃 
Thanks for the fast review!

libcraft/core/src/dimension.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    pub fn namespaced_id(&self) -> &'static str {
+        match self {
+            Self::Overworld => "minecraft:overworld",
+            Self::TheNether => "minecraft:the_nether",
+            Self::TheEnd => "minecraft:the_end",
+        }
+    }
+
+    pub fn from_namespaced_id(id: &String) -> Option<Self> {
```
Defman, line 38: Should be `&str`

libcraft/core/src/dimension.rs
```diff
@@ -0,0 +1,68 @@
...
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(try_from = "String", into = "&'static str")]
+pub enum Dimension {
+    Overworld,
+    TheNether,
+    TheEnd,
+}
+
+impl Dimension {
+    pub fn dim_id(&self) -> i32 {
```
Defman, line 13: Should maybe implement `From<i32> for Dimession` and `From<Dimension> for i32` in addition.

libcraft/core/src/dimension.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub fn from_dim_id(id: i32) -> Option<Self> {
+        match id {
+            0 => Some(Self::Overworld),
+            -1 => Some(Self::TheNether),
+            1 => Some(Self::TheEnd),
+            _ => None,
+        }
+    }
+
+    pub fn namespaced_id(&self) -> &'static str {
```
Defman, line 30: Maybe `From<Dimension> for &'static str` in addition.

Defman (changes requested):

libcraft/core/src/dimension.rs
PauMAVA: He needs `TryFrom` in order to use `#[serde(try_from = "String", into = "&'static str")]` right?

libcraft/core/src/dimension.rs
PauMAVA: Same as thing as I commented in `TryFrom`?

PauMAVA:
> 
> 
> Not sure how I feel about the name "namespaced_id". We call it "name" everywhere else, but I think namespaced id makes more sense. But I do think that consistency is VERY important. So we need to make a choice here. What do you think @Defman

Well, I think you are right. Here we have the problem of consistency and the problem of misleading names. `name` is a completely different term than `namespaced_id`. `name` is more like a `DisplayName` or `PlayerName` imo.

libcraft/core/src/dimension.rs
Defman: True, forgot about that. There is an open PR for `serde` to add `from_str`.

Defman:
```rust
pub trait NameSpace {
  // Possible `Cow<'static, str>`, if not all namespaces is know at compile time
  fn name(&self) -> &'static str;
  fn from_name(name: &str) -> Self;
}
```

libcraft/core/src/dimension.rs
Defman: No its not the same, he should implement `From<Dimension> for &'static str`, this will automaticlly implement `Into<&'static str> for Dimension`

PauMAVA: Oh didn't know that. So implementing `From` does implement `Into` automatically?

Defman: One should always prefer implementing `From` over `Into` because implementing `From` automatically provides one with an implementation of `Into` thanks to the blanket implementation in the standard library.

https://doc.rust-lang.org/stable/std/convert/trait.From.html

libcraft/core/src/dimension.rs
Masplus: I think it should be `TryFrom<i32> for Dimension` as the conversion can fail.

Defman: Ahh yes, indeed it should be.

libcraft/core/src/dimension.rs
PauMAVA: Okay, thanks! 😄

libcraft/core/src/dimension.rs
Defman: Also this is merely a suggestion, I'm not sure if we should implement `From` or `TryFrom<i32>` I'm leaning towards not, just wanted to bring up the possibility.

PauMAVA: Well, he has implemented it already so we will have the possibility xD.

Defman (approved):
I think we should revisit the idea of a general `NameSpace` trait in the future. For now we will merge and we can open an issue regarding consistency of through a shared `NameSpace` trait.

PauMAVA:
> 
> 
> I think we should revisit the idea of a general `NameSpace` trait in the future. For now we will merge and we can open an issue regarding consistency of through a shared `NameSpace` trait.

Sounds good!

PauMAVA (approved):
LGTM!

libcraft/core/src/dimension.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+impl From<Dimension> for &'static str {
+    fn from(value: Dimension) -> Self {
+        value.namespaced_id()
+    }
+}
+
+impl TryFrom<i32> for Dimension {
```
Schuwi, line 66: I don't think we should have From/Into conversions for numerical IDs as their use is discouraged but if there is a good reason please tell me.

libcraft/core/src/dimension.rs
```diff
@@ -0,0 +1,82 @@
...
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(try_from = "String", into = "&'static str")]
+pub enum Dimension {
+    Overworld,
+    TheNether,
+    TheEnd,
+}
+
+impl Dimension {
+    pub fn dim_id(&self) -> i32 {
```
Schuwi, line 13: To maintain consistency in libcore this should be renamed `id` and `from_id`, see for example https://github.com/feather-rs/feather/blob/main/libcraft/particles/src/particle.rs or https://github.com/feather-rs/feather/blob/main/libcraft/core/src/biome.rs.

Schuwi (commented):
Thanks for your PR!

libcraft/core/src/dimension.rs
PauMAVA: Why the use of `From/Into` for numerical IDs is discouraged?

Schuwi: I meant that the use of numerical IDs in general is discouraged by Mojang and thus I am not sure a convenience function like From/Into makes sense to implement

PauMAVA: Oh, I see. That's true but I don't actually see any problem in implementing `Into/From` for them as despite being discouraged they are not forbidden thus they can be used safely by the developer imo. Also, I think @Defman suggested @Masplus add this in a prior review of his.

Masplus: @Defman What's your opinion about this?

Defman: I did comment on this earlier.
```
Also this is merely a suggestion, I'm not sure if we should implement From or TryFrom<i32> I'm leaning towards not, just wanted to bring up the possibility.
```
The reason why I'm leaning towards not, is because an `i32` is not very descriptive and the conversion should not happen for all `i32`. Also I agree with @Schuwi that we should prefer namespace id's over ids.

I suggest
1. Remove the `TryFrom`
2. Implement the `NameSpace` trait as discussed
3. Maybe Implement `Numerical` trait with `as_numerical(&self) -> i32` and `from_numerical(i32) -> Option<Self>`.

I'm not sold on the `Numerical` trait.

PauMAVA: I think the `NameSpace` thing should be discussed in another issue and have its own PR so that we will be able to migrate from simple `str` to `impl NameSpace` on all types that use it. I can open that issue for you now if you are ok with it.

Masplus:
I think this is ready for merging.

Defman (approved):
Thanks, looks good to me.

amberkowalski (approved):
Looks good for feather in its current state.

PR #390, opened 2021-03-08
Text refactoring

Defman:
Refactors the text API and adds ANSI terminal support.

There is still a few things missing:
- Serialization and Deserialization, not quite sure how it should be handled, maybe an IR enum.
- Allow translation to work server-side
- Scoreboard (similar to translation)
- Keybind (similar to translation)


feather/common/src/chat.rs
```diff
@@ -51,21 +51,13 @@ impl ChatBox {
 /// Represents a chat message.
 #[derive(Debug, Clone)]
 pub struct ChatMessage {
-    kind: ChatKind,
-    message: Text,
+    pub kind: ChatKind,
```
caelunshun, line 54: These should be kept provide for encapsulation.

caelunshun (commented):
Looks good so far, thanks!

PR #394, opened 2021-03-08
Health System

### LINKED ISSUE #358: Health and Death
Reimplement health and death. This feature existed before #307 but was lost in the rewrite. You can refer to the `old` directory for how it was previously implemented.

In short, we want:
* A `Health` component containing an entity's health and its max health, with methods to deal damage and regenerate.
* A system that queries for all entities with `Health` components and removes the entities if `health == 0`. For players, we instead want to respawn them, which is trickier. It involves sending the [`ChangeGameState` packet](https://wiki.vg/Protocol#Change_Game_State) and waiting for the client to send [`ClientStatus`](https://wiki.vg/Protocol#Client_Status). The hard part here is that the player should not be visible to the game while they are in the respawn menu. I'm not sure what the best way to implement this is.
* To test health, implement fall damage. This entails a system that checks the `Position` and `OnGround` components each tick and computes how far an entity has fallen.

kirawi:
# Health System

## Status

- [ ] Ready 
- [x] Development
- [ ] Hold

## Description

Implements systems and events for handling updates to an entity's health, as well as systems that handle a player's hunger.

## Related issues

Resolves #358 

## Checklist

- [ ] Ran `cargo fmt`, `cargo clippy`, `cargo build --release` and `cargo test` and fixed any generated errors!

Will be fleshed out with details once ready for review.


feather/common/src/events.rs
```diff
@@ -8,7 +8,10 @@ use crate::view::View;
 mod block_change;
 mod plugin_message;
 
+pub mod health;
+
 pub use block_change::BlockChangeEvent;
+pub use health::HealthEventType;
 pub use plugin_message::PluginMessageEvent;
 
 /// Triggered when a player joins the `Game`.
```
amberkowalski, line 17: Might be better to just use game without the \` since its not a type, term, or variable name.

feather/server/src/systems/health.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    // if game.tick_count % 8 == 0 {
+    //     for (player, (client_id, health)) in game.ecs.query::<(&ClientId, &mut Health)>().iter() {
+    //         if let Some(client) = server.clients.get(*client_id) {
+    //             health.deal_damage(1);
+    //             client.update_health(&health);
+    //         }
+    //     }
+    // }
```
amberkowalski, line 35: Commented out code block.

quill/common/src/components.rs
```diff
@@ -98,3 +98,53 @@ impl Display for CustomName {
...
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Health {
+    pub health: u32,
+    pub max_health: u32,
+}
+
+impl Health {
+    pub fn new(max_health: u32) -> Self {
+        let health = 1;
```
amberkowalski, line 110: I think the default should be full health

amberkowalski (changes requested):
I overall am a little worried about using constants, but I'm not sure to explain what I mean.

A few questions:
1. How would this work with creative mode
    - How would this work with void damage in creative mode.
2. If you were a server owner who wanted to disable hunger, could you do it easily?
3. What if the no regen gamerule was on?


These arent meant to quiz your solution, just questions to think about to ensure flexibility.

quill/common/src/components.rs
kirawi: Sorry, a lot of this is just placeholder code as I test that everything is working!

feather/common/src/events.rs
amberkowalski: The backticks around "Game"

kirawi: Oh, that's not my change.

amberkowalski: Welp, might as well fix it while we're here I guess.

kirawi:
Thoughts on the events? I decided to put the ones Feather needed inside a `mod` as events inside `quill`, so it should be accessible to plugins, I think. People can still make their own custom ones if they want.

feather/protocol/src/packets/server/play.rs
```diff
@@ -1179,6 +1181,163 @@ packets! {
...
+            EntityAttributeKind::Armor => 30.0,
+            EntityAttributeKind::ArmorToughness => 20.0,
+            EntityAttributeKind::Luck => 1024.0,
+            EntityAttributeKind::HorseJumpStrength => 2.0,
+            EntityAttributeKind::ZombieSpawnReinforcements => 1.0,
+        }
+    }
+
+    pub fn min_value(&self) -> f32 {
+        if let EntityAttributeKind::Luck = &self {
```
Schuwi, line 1266: I would actually rather make this a match as well because it makes more semantic sense imo.
```rust
match &self {
    EntityAttributeKind::Luck => -1024.0,
    _ => 0.0,
}
```

feather/protocol/src/packets/server/play.rs
```diff
@@ -1179,6 +1181,163 @@ packets! {
...
+            EntityAttributeKind::AttackSpeed => "generic:attack_speed",
+            EntityAttributeKind::AttackKnockback => "generic:attack_knockback",
+            EntityAttributeKind::FlyingSpeed => "generic:flying_speed",
+            EntityAttributeKind::Armor => "generic:armor",
+            EntityAttributeKind::ArmorToughness => "generic:armor_toughness",
+            EntityAttributeKind::Luck => "generic:luck",
+            EntityAttributeKind::HorseJumpStrength => "horse:jump_strength",
+            EntityAttributeKind::ZombieSpawnReinforcements => "zombie:spawn_reinforcements",
+        })
+    }
```
Schuwi, line 1245: Replace those by defining `EntityAttributeKind` in a `def_enum` if possible.

feather/protocol/src/packets/server/play.rs
kirawi: I think clippy complains about that.

feather/protocol/src/packets/server/play.rs
```diff
@@ -1179,6 +1181,163 @@ packets! {
...
+    pub amount: f32,
+    pub operation: OperationKind,
+}
+
+#[derive(Debug, Clone)]
+pub struct EntityProperty {
+    pub attribute: EntityAttributeKind,
+    pub modifiers: Vec<AttributeModifier>,
+    pub value: f32,
+}
```
Schuwi, line 1286: I think you just define those in a `packets!` macro. See for example https://github.com/feather-rs/feather/blob/a9bb1e639c2addc2fc736a68f4a01e94e91d5c24/feather/protocol/src/packets/server/play.rs#L797-L800

feather/protocol/src/packets/server/play.rs
```diff
@@ -1179,6 +1181,163 @@ packets! {
...
+pub struct AttributeModifier {
+    pub uuid: Uuid,
+    pub amount: f32,
+    pub operation: OperationKind,
+}
+
+#[derive(Debug, Clone)]
+pub struct EntityProperty {
+    pub attribute: EntityAttributeKind,
+    pub modifiers: Vec<AttributeModifier>,
```
Schuwi, line 1284: You need to define the type of protocol vec you need here as well. As the length is given by a `VarInt` here according to wiki.vg you should use a `LengthPrefixedVec`.

feather/protocol/src/packets/server/play.rs
```diff
@@ -1179,6 +1181,163 @@ packets! {
...
+        for modifier in &self.modifiers {
+            modifier.uuid.write(buffer, version);
+            modifier.amount.write(buffer, version);
+            modifier.operation.write(buffer, version);
+        }
+
+        println!("{:?}", buffer);
+    }
+}
+
```
Schuwi, line 1340: I didn't check these implementations as they should be unnecessary once you define `AttributeModifier`, `EntityProperty` and `EntityAttributeKind` with macros (See my other comments).

feather/protocol/src/packets/server/play.rs
```diff
@@ -1179,6 +1181,163 @@ packets! {
...
+            EntityAttributeKind::MovementSpeed => "generic:movement_speed",
+            EntityAttributeKind::AttackDamage => "generic:attack_damage",
+            EntityAttributeKind::AttackSpeed => "generic:attack_speed",
+            EntityAttributeKind::AttackKnockback => "generic:attack_knockback",
+            EntityAttributeKind::FlyingSpeed => "generic:flying_speed",
+            EntityAttributeKind::Armor => "generic:armor",
+            EntityAttributeKind::ArmorToughness => "generic:armor_toughness",
+            EntityAttributeKind::Luck => "generic:luck",
+            EntityAttributeKind::HorseJumpStrength => "horse:jump_strength",
+            EntityAttributeKind::ZombieSpawnReinforcements => "zombie:spawn_reinforcements",
```
Schuwi, line 1243: According to wiki.vg all of these identifiers are separated by `.` not by `:`. E.g. `generic.max_health` instead of `generic:max_health`.

kirawi: In https://wiki.vg/Protocol#Statistics , it mentions that they separate with `.` but it's actually `:`. I'm going with that assumption here as well, but I'll test it later.

feather/protocol/src/packets/server/play.rs
```diff
@@ -1179,6 +1181,163 @@ packets! {
...
+    pub uuid: Uuid,
+    pub amount: f32,
+    pub operation: OperationKind,
+}
+
+#[derive(Debug, Clone)]
+pub struct EntityProperty {
+    pub attribute: EntityAttributeKind,
+    pub modifiers: Vec<AttributeModifier>,
+    pub value: f32,
```
Schuwi, line 1285: This needs to be between `attribute` and `modifiers` when you implement this struct via the `packets!` macro because the declaration order needs to match the protocol order in that case.

Schuwi, line 1285: wiki.vg specifies a type of `Double` here which is a `f64` in Rust.

feather/protocol/src/packets/server/play.rs
```diff
@@ -1179,6 +1181,163 @@ packets! {
...
+        }
+
+        0.0
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct AttributeModifier {
+    pub uuid: Uuid,
+    pub amount: f32,
```
Schuwi, line 1277: Just like `EntityProperty.value` this has to be a `f64`.

feather/protocol/src/packets/server/play.rs
```diff
@@ -1162,7 +1162,9 @@ packets! {
     }
 
     EntityProperties {
-        __todo__ LengthInferredVecU8;
+        entity_id VarInt;
+        properties LengthPrefixedVec<EntityProperty>;
```
Schuwi, line 1166: This needs to be an `IntPrefixedVec` which you need to implement first, like described on Discord.
Kinda forgot to comment that here and it might not have been that clear in Discord what I mean specifically.

kirawi:
Depends upon #410

kirawi:
Seems like the only things left to be implemented are the hunger system, handling entity properties, updating relevant entity metadata, and fall damage. Maybe I'll implement an example plugin as well. That should be it for an initial health system.

kirawi:
This is going to be entirely rewritten from the ground up, it would be harder to use this PR to track it.

Issue #402, opened 2021-03-12
Proposal: New ECS implementation + bundles and accessors


caelunshun:
In implementing custom components for plugins, I realized that the ECS implementation needs to allocate custom plugin components within the plugin's linear memory. We thus have two options:
* continue using a further modified fork of `hecs`; or
* implement our own ECS.

I am inclined to proceed with the latter because a custom ECS designed to work with the plugin system will be more flexible. Also, I have some ideas to solve #397 that we can integrate into the new ECS.

### New ECS design

This is my plan for the new ECS:
* It will be based on sparse sets, like [EnTT](https://github.com/skypjack/entt), whereas `hecs` uses archetypes. Sparse sets are easier and safer to implement than archetypes, and they have better performance characteristics when inserting/deleting components at a high frequency. (Our component-based event system does a lot of insertions and deletions, which convinces me that sparse sets are the best choice for Feather.) 
* It will support more flexibility with the borrow checker. In particular, components will have a stable location in memory, even if new entities and components are created. As a result, it should be possible to add/remove components and spawn entities inside a query loop, which will be convenient for some use cases in Quill.
* The `Component` trait will not be implemented for all `T`; each component type needs to implement the trait instead. While this is less convenient in some areas, it prevents querying for types that are not components, like bundles. (We'll provide a derive macro for `Component` to avoid too much boilerplate.)
* It will have built-in support for _bundles_ and _bundle accessors_ as an alternative implementation to #388 and #397. My vision is this: each entity type has its own `Bundle` struct, which is the set of components it spawns with. The ECS crate provides a derive macro that generates an _accessor_ struct for the bundle that can be queried. In code, this looks like:

```rust
#[derive(Bundle)]
#[bundle(generate_accessor = "Player")] // generates the Player struct
pub struct PlayerBundle {
    /// another bundle containing default components for all living entities, like `Position` and `Health`
    /// (this emulates inheritance via composition)
    #[bundle(inherit)]
    base: LivingEntityBundle,
    inventory: Inventory,
    player_marker: PlayerMarker,
}

impl Player {
    pub fn item_in_main_hand(&self) -> &ItemStack { /* ... */ }
}

// To spawn a player:
game.spawn_entity(PlayerBundle {
    base: LivingEntityBundle {
        pos,
        ..Default::default()
    },
    ..Default::default()
});

// The accessor for the bundle can be queried using a new `query_as` function
// to distinguish accessors from real components.
[...]

PR #404, opened 2021-03-13
Implement new custom ECS (WIP)

caelunshun:
## Status

- [ ] Ready 
- [x] Development
- [ ] Hold

## Description

Implements #402. Work in progress.


quill/ecs/src/component.rs
```diff
@@ -0,0 +1,67 @@
...
+/// "opaque" types identified by a `u64`.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct ComponentTypeId(ComponentTypeIdInner);
+
+impl ComponentTypeId {
+    pub fn of<T: 'static>() -> Self {
+        Self(ComponentTypeIdInner::Rust(TypeId::of::<T>()))
+    }
+
+    pub fn opaque(id: u64) -> Self {
```
Defman, line 17: Maybe 32 bits would be enough?

quill/ecs/src/component.rs
```diff
@@ -0,0 +1,67 @@
...
+        Self(ComponentTypeIdInner::Opaque(id))
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+enum ComponentTypeIdInner {
+    Rust(TypeId),
```
Defman, line 24: Maybe just use `Opaque` and use a `FnvHashMap` to map from type id to opaque id.

quill/ecs/src/storage/blob_vec.rs
```diff
@@ -0,0 +1,289 @@
...
+        }
+
+        self.check_invariants();
+    }
+
+    fn grow(&mut self) {
+        let old_capacity = self.capacity;
+        let new_capacity = match old_capacity {
+            0 => 4,
+            x => x.checked_mul(2).expect("capacity overflow"),
```
Defman, line 72: `next_power_of_two` could maybe be a good alternative?

quill/ecs/src/storage/sparse_set.rs
```diff
@@ -0,0 +1,118 @@
+use std::{iter, mem::MaybeUninit, ptr::NonNull};
+
+use component::ComponentTypeId;
+
+use crate::component::{self, ComponentMeta};
+
+use super::blob_vec::BlobVec;
+
+/// Stores components in a sparse set.
+pub struct SparseSetStorage {
```
Defman, line 10: `FnvHashMap` is insanely fast, maybe it could be used in stead of `SparseSetStorage`?

Might require some architectural changes, since for each entity you would lookup its components by their id.

quill/ecs/src/entity.rs
```diff
@@ -0,0 +1,25 @@
...
+        let generation = bits as u32;
+        Self { index, generation }
+    }
+
+    pub fn index(self) -> u32 {
+        self.index
+    }
+
+    pub fn generation(self) -> u32 {
+        self.generation
```
Defman, line 23: What is `generation` used for, can't quite seem to connect it.

Defman (commented):
Looks good, some minor questions.

Could we perhaps use `FnvMap` for component storage since it is insanely fast. I know we are not using an ArchType based system, but my example is based on it.

```rust
struct World {
  // Finding the `ArchTypeId` for a given set of components, 
  // used when spawning a new entity or adding a component to an existing entity.
  // Note btree maps and sets might not be ideal, since they are kinda slow.
  arch_type_sets: BTreeMap<BTreeSet<ComponentId>, ArchTypeId>,
  // When quering we and all components bitset together, 
  // which yields indices into `arch_types`.
  components: FnvHashMap<ComponentId, BitSet>,
  arch_types: Vec<ArchType>,
  // Mapping from entity id to memory location.
  entities: FnvHashMap<EntityId, (ArchTypeId, ArchTypeOffset)>,
}

struct ArchType {
  components: FnvHashMap<ComponentId, MemorySpace>,
}
```

quill/ecs/src/component.rs
amberkowalski: There's no harm in using a u64 so I disagree

quill/ecs/src/storage/sparse_set.rs
caelunshun: I'm using sparse sets instead of hash tables because we can optimize query iteration by e.g. packing all entities matching a given archetype into the front of the array. See https://skypjack.github.io/2019-03-07-ecs-baf-part-2/

quill/ecs/src/entity.rs
caelunshun: It's used to solve the ABA problem, since entity indices are reused. A newly created entity could have the same index as an old, despawned entity, so we use the generation to ensure old `Entity` handles don't refer to new entities.

quill/ecs/src/component.rs
Schuwi: How would that be better?

Defman: We have to chose a true id at one point, I suggest we chose `Opaque` as the true id and map `TypeId` to the true id(`Opaque`). This indirection have happen at some point.

quill/ecs/src/storage/blob_vec.rs
Schuwi: I would assume the `next_power_of_two` function is more complex than a simple multiplication and I doubt the compiler can elide that extra logic because it can't make assumptions about the values of `self`. What would be the benefit?

Defman: I'm not saying there is anything wrong with this implementation, I'm only suggesting `next_power_of_two` since that is how `RawVec`, by extension `Vec`, is implemented.

amberkowalski: Actually from what I can see the next_power_of_two function just does bit shifting.

quill/ecs/src/storage/sparse_set.rs
Schuwi: Wouldn't using a HashMap defeat part of the purpose of ECS by getting rid of linear memory reads?

Defman: Sparse set already gets rid of 100% linear memory reads. Also `FnvHashMap` is not just fast, its insanely fast... it even beats bitsets... its crazy.

Schuwi: I mean the memory reads with sparse sets are still linear, just with gaps.
Using a HashMap will make memory reads completely unpredictable.

quill/ecs/src/component.rs
Schuwi: What do you mean by "true id" and why is it needed?

quill/ecs/src/storage/blob_vec.rs
Schuwi: I cannot find a call to `next_power_of_two` anywhere in `raw_vec.rs`...
`grow_amortized` uses a simple `* 2` multiplication.
Also, looking at the implementation `next_power_of_two` is noticeably more complex than a simple multiplication.

amberkowalski: I've been lied to.

quill/ecs/src/storage/sparse_set.rs
Defman: Hmm now I'm wondering where I saw `next_power_of_two` being used. `RawVec` uses the following algorithm
```rust
let required_cap = used_cap + needed_extra_cap;
let double_cap = self.cap * 2;
let new_cap = cmp::max(required_cap, double_cap);
```

quill/ecs/src/entity.rs
Defman: Another approach could also be to queue up insertions and deletions of components and entities. And when the `Query` is dropped, commit the queue to the `World`.

quill/ecs/src/storage/sparse_set.rs
Schuwi: These two code snippets produce different behaviour. Also you are in the wrong comment thread xD

quill/ecs/src/entity.rs
Schuwi: That would not let us get rid of generations

quill/ecs/src/storage/sparse_set.rs
Defman: Everything just seems to be going wrong for me at the moment 😅

Yea, the behavior is slightly different but both ensures push are `O(n)` amortized. The implementation that does not use `next_power_two` will double the capacity unless the total required cap is greater than the doubled size. Where the `next_power_of_two` will double and if the capacity is greater than `cap * 2` it will be `cap * 2 * 2` it will repeat this until double cap is equal or greater than required cap.

quill/ecs/src/entity.rs
Defman: True an old `Entity` id could be pointing to a new `Entity`, should we allow end users to store long lived entity ids?

caelunshun: Yes, storing long-lived entities is definitely needed. Feather uses it for the `ChunkEntities` data structure, for example.

quill/ecs/src/component.rs
Defman: "true id" one universal identifier for components, no distinction between local ie `Any::TypeId` or remote ids.

quill/ecs/src/entity_builder.rs
```diff
@@ -0,0 +1,123 @@
...
+impl EntityBuilder {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Inserts a new component for the entity.
+    ///
+    /// If the entity builder already contains the component,
+    /// then the previous value is overriden.
+    pub fn add<T: Component>(&mut self, component: T) -> &mut Self {
```
Defman, line 27: `Component` does not require `Copy` which means `T` could contain ptrs which will be dropped at the end of add?

caelunshun: We wrap `component` in a `MaybeUninit` to prevent it from being dropped.

amberkowalski: Components are bound on Sync.

caelunshun: Only `Send`, not `Sync`, right now. `Ecs` doesn't implement `Sync` as a result.

amberkowalski: I typed it out before checking and saw `Send` and sent the comment. But the idea behind it is still correct.

caelunshun: He's referring to structs like `Vec` which hold pointers - if they're dropped twice, it's UB because of a double free.

Defman: We still have to call `drop` on clear `EntityBuilder::clear` this was my roundabout way of getting there =)

quill/ecs/src/component.rs
Schuwi: And **why is it needed**?

caelunshun:
I updated the PR with a checklist of remaining work.

Defman:
BUMP

Defman:
Bump

Issue #329, opened 2020-12-29
Console commands?


XfedeX:
How to execute console commands??
Console don't seem to have commands support....

PR #407, opened 2021-03-25
Commands, they said it would be a good first issue. 

Miro-Andrin:
# Regex backed command dispatching.
## Status

- [ ] Ready
- [x] Development
- [ ] Hold

## Description
I added rudimentary command dispatching, with an example plugin in quill/example-plugins/command. The example shows how to create a command "/echo <u32>", that prints a message to the caller.

This pull request should not be accepted until my pull request to lieutenant has been accepted. Currently, it has my lieutenant branch/repo set as a dependency in Cargo.toml.

### Limitations:
1) It only works for chat messages at the moment, but it would not be difficult for someone (or maybe me) to add rudimentary console command dispatching.  #329.  We could also consider adding a way for plugins to call each other.

2) It only allows for adding commands, so if someone were to load and unload a plugin there would be some build-up of garbage  in the command dispatcher. I or someone else just needs to hook into when a plugin is unloaded, and then  update the list of commands in  the command dispatcher.

###  Improvements I am currently working on.
1) Better error message when command registration fails. I am currently looking into this, by adding some functionality in lieutenant to detect when there is a regex using unsupported features.  So instead of just panicking
with a "could not load command", it should give an example of the regex feature casing the error.

2) Basic Diagnostics. We can get some interesting diagnostics from the command dispatcher. To begin with  i would just like to know.
a) How many commands are loaded.
b) How much memory the DFA is using,
c) Number of commands that can't be distinguished based on their regex. 

## Related issues
 #329, #229

## Checklist
- [ x] Ran `cargo fmt`, `cargo clippy`, `cargo build --release` and `cargo test` and fixed any generated errors!
- [ x] Removed unnecessary commented out code

### DIFF
--- a/feather/plugin-host/src/command_dispatcher.rs
+++ b/feather/plugin-host/src/command_dispatcher.rs
@@ -5,11 +5,12 @@ use std::{
     rc::Rc,
 };
 
+use crate::{context::PluginPtrMut, Game, PluginId, PluginManager};
 use anyhow::bail;
 use feather_ecs::{Entity, EntityRef};
 use lieutenant::regex::early_termination::CmdPos;
 use lieutenant::regex::{dfa::DFA, NFA};
-use crate::{context::PluginPtrMut, Game, PluginId, PluginManager};
+use quill_common::EntityId;
 
 // Constant that should reflect how many commands a default
 // server starts with. Vanilla has roughly 60
@@ -17,8 +18,8 @@ const NUM_COMMANDS: usize = 20;
 
 type CommandId = u32;
 
-/// A collection of data containing what plugin a command was defined in, 
-/// the pointer to its Box<Command<...>>, and the regex that determines 
+/// A collection of data containing what plugin a command was defined in,
+/// the pointer to its Box<Command<...>>, and the regex that determines
 /// when the command should be called.(The command does some parsing beyond the regex).
 pub struct CommandDescription {
     pub plugin_id: PluginId,
@@ -32,19 +33,21 @@ impl CommandDescription {
     // Returns true or false depending on if the command executed or not.
     // If it returns false that means the commands parser did not recognize
     // the commnd.
-    fn call(&self, game: &mut Game, input: &str) -> anyhow::Result<()> {
+    /// caller: None => caller was terminal,
+    /// caller: Some(x) => player x called command.
+    fn call(&self, game: &mut Game, input: &str, caller: Option<EntityId>) -> anyhow::Result<()> {
         let plugin_manager = Rc::clone(&*game.resources.get::<Rc<RefCell<PluginManager>>>()?);
         let plugin_manager = plugin_manager.borrow();
         let plugin = plugin_manager.plugin(self.plugin_id);
 
         match plugin {
             Some(plugin) => {
-                plugin.call_command(game, input, self.callback)?;
+                plugin.call_command(game, input, self.callback, caller)?;
                 Ok(())
             }
             None => {
                 // Plugin has been unloaded.
-                bail!("Plugin was unloaded.")
+                bail!("Plugin was not loaded.")
             }
         }
     }
@@ -75,12 +78,12 @@ impl CommandDispatcher {
         // This method should not be called for every command we
         // add. It is somewhat expensive. It is O(n) in the number of
         // commands we have registerd so far. Therefor adding m commands
-        // is a O(m^2) operation. 
-        // Instead we should (option a) try to build only the dfa for the 
+        // is a O(m^2) operation.
+        // Instead we should (option a) try to build only the dfa for the
         // single cmd (to check that its regex does not use any unimplemented features)
         /// by calling  NFA::<CmdPos<CommandId>>::from_command_regex(&cmd.regex,0).is_ok();
         /// or (option b) add a method to lieutenants NFA that checks if the regex
-        /// contains any unimplemented features. See lieutenant/regex/regex_to_nfa, and 
+        /// contains any unimplemented features. See lieutenant/regex/regex_to_nfa, and
         /// call that method instead. Anyways for these options we need someway of triggering
         /// dfa updates. Could probably happen inside plugin load function, in main, or a host-call.
         self.update_dfa();
@@ -104,12 +107,12 @@ impl CommandDispatcher {
     }
 
     /// calls the corresponding command (if it exists) and returns the resulting u32
-    /// player is person calling the command, None means terminal.
+    /// Caller is player calling the command, None means terminal.
     pub fn call(
         &self,
         game: &mut Game,
         input: &str,
-        player: Option<Entity>,
+        caller: Option<EntityId>,
     ) -> anyhow::Result<i64> {
         match self.dfa.early_termination_find(input) {
             Ok(ids) => {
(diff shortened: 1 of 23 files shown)


ColdIce1605:
Any more progress?

Miro-Andrin:
Yes. I have rewritten the regex stuff such that there is no difference between nfa and dfa, and instead i use a hybrid data structure. This way of doing things is much better, because it resulted in less code, and allows the dispatcher to be lazier and smarter about when to do the time vs space optimisations.  I also have working tab completion based on a commands regex. It uses the dumbest, most straight forward way of doing it, and is probably not fast.  

Another thing i have been looking at is how to improve the command building code. Currently we have a builder pattern like:
`
let cmd = literal("/add")
.space()
.arg<i64>()
.space()
.arg<i64>()
.on_call(|x:i64,y:i64| -> move |game: Game, caller: Caller| Ok(x+y)});
`
I don't like how horrible the error messages are when the types don't align. I have tried a couple of things like using macros, but nothing worked. Any ideas on improving this are welcomed. 

I have cleaned up how optional arguments are defined, so now you can write arg<Option<i64>> and it works like expected. You don't need to do .opt_arg<i64>. However by doing that the .space() no longer works. So there is a implicit one or more space between every argument. I think this is a improvement.

Defman:
@Miro-Andrin what's left, before this is in a state where it can be merged?

feather/plugin-host/src/command_dispatcher.rs
```diff
@@ -0,0 +1,163 @@
...
+use feather_ecs::{Entity, EntityRef};
+use lieutenant::regex::early_termination::CmdPos;
+use lieutenant::regex::{dfa::DFA, NFA};
+use quill_common::EntityId;
+
+// Constant that should reflect how many commands a default
+// server starts with. Vanilla has roughly 60
+const NUM_COMMANDS: usize = 20;
+
+type CommandId = u32;
```
romangraef, line 19: I am not sure if this will end up being a good idea.

feather/plugin-host/src/command_dispatcher.rs
```diff
@@ -0,0 +1,163 @@
...
+
+/// A collection of data containing what plugin a command was defined in,
+/// the pointer to its Box<Command<...>>, and the regex that determines
+/// when the command should be called.(The command does some parsing beyond the regex).
+pub struct CommandDescription {
+    pub plugin_id: PluginId,
+    pub regex: String,
```
romangraef, line 26: Why are we using a regex in here, if we are already using lieutenant?

feather/plugin-host/src/command_dispatcher.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub callback: PluginPtrMut<u8>,
+}
+
+impl CommandDescription {
+    // Returns true or false depending on if the command executed or not.
+    // If it returns false that means the commands parser did not recognize
+    // the commnd.
+    /// caller: None => caller was terminal,
+    /// caller: Some(x) => player x called command.
+    fn call(&self, game: &mut Game, input: &str, caller: Option<EntityId>) -> anyhow::Result<()> {
```
romangraef, line 38: I personally dislike the idea of having an `Option` to represent whether the terminal or a player was the command sender, as it isn't directly intuitive, and isn't necessarily future-proof if we also consider command blocks. On top of that it might be nice to have a general trait for replying or querying permissions, which could simplify some applications.

feather/plugin-host/src/command_dispatcher.rs
```diff
@@ -0,0 +1,163 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.dfa.fmt(f)
+    }
+}
+
+impl CommandDispatcher {
+    pub fn new() -> Self {
+        Self {
+            dfa: DFA::new(),
+            commands: Vec::with_capacity(NUM_COMMANDS),
```
romangraef, line 71: I'm again skeptical of having this as a sort of magic constant, then again this is just the initial capacity, so it's probably fine.

feather/plugin-host/src/command_dispatcher.rs
```diff
@@ -0,0 +1,163 @@
...
+
+        for (id, cmd) in self.commands.iter().enumerate() {
+            let cmd_nfa = NFA::<CmdPos<CommandId>>::from_command_regex(
+                &cmd.regex,
+                id.try_into()
+                    .expect("Cant register 2^32 different commands!"),
+            )?;
+            nfa = nfa.or(cmd_nfa)?;
+        }
+        // Reduses the number of states/nodes, so it cant fail.
```
romangraef, line 104:
```suggestion
        // Reduces the number of states/nodes, so it can't fail.
```

feather/plugin-host/src/command_dispatcher.rs
```diff
@@ -0,0 +1,163 @@
...
+                            ) {
+                                Ok(x) => {
+                                    // Command sucsessfully ran
+                                    return Ok(x);
+                                }
+                                Err(_) => {
+                                    // Was not able to call command, probably due to failed parsing.
+                                    // Unless there is some failure case i have not considerd, or someone
+                                    // is not using lieutenant, but their own shenanigans.
+                                    continue;
```
romangraef, line 143: Maybe we could have custom `bail!`s for those cases as well, instead of falling back to the default return.

feather/server/src/packet_handlers.rs
```diff
@@ -123,10 +127,34 @@ fn handle_animation(
...
+        packet.message.as_str(),
+        Some(EntityId(player_id.to_bits())),
+    ) {
+        Err(_) => {
+            // Command did not run because parsing failed, or len ofmessage was longer then u32::MAX
+            let player = game.ecs.entity(player_id)?;
+            let name = player.get::<Name>()?;
+            let message =
+                Text::translate_with("chat.type.text", vec![name.to_string(), packet.message]);
+            game.broadcast_chat(ChatKind::PlayerChat, message);
```
romangraef, line 150: I don't think this is an appropriate way too handle chat messages, since regular chat messages without a prefixed `/` should not result in an error message, and messages that result in an erroneous command execution shouldn't be broadcasted either. A more appropriate way too handle this, would be to first check for a prefix `/` and then handle eventual command execution errors.

feather/plugin-host/src/command_dispatcher.rs
Defman: Llieutenant internally uses a custom Regex engine, this just passes the Regex from client(plugin) to host(server) such the server can ask the client(plugin) to do the final matching.

feather/plugin-host/src/command_dispatcher.rs
Defman: I agree, @Miro-Andrin therterminal is represented in the ECS and therefore also has an `EntityId`.

feather/plugin-host/src/command_dispatcher.rs
Defman: The default capacity of `Vec` should be fine.

feather/plugin-host/src/command_dispatcher.rs
romangraef: Ah thanks for the clarification, i only used regular (java) brigadier :)

feather/server/src/packet_handlers.rs
Miro-Andrin: This actually does not print a error message when a error happens. Notice how the error is never used.  Its the opposite way around. It does not print anything if a command succeeds.  So yeah, if you write "/not a actual command", then  "/not a actual command" appears in the chat.  This is like you noted not the correct behaviour.  
I am going to change this code to be clearer.

feather/plugin-host/src/command_dispatcher.rs
Miro-Andrin: 100% Agree, that is just me being lazy.

Defman:
BUMP

PR #410, opened 2021-04-09
Make LengthPrefixedVec generic

kirawi:
# Make LengthPrefixedVec generic

## Status

- [x] Ready

## Description

Made it generic and added typedefs for compatibility with existing code, and removed `ShortPrefixedVec` implementation. As per this [conversation](https://discord.com/channels/619316022800809995/805342116057579520/829511984423632916). Honestly, I'm not sure if performance will suffer, but I hope not.

## Checklist

### DIFF
--- a/feather/protocol/src/io.rs
+++ b/feather/protocol/src/io.rs
@@ -299,7 +299,7 @@ impl Writeable for VarLong {
                 break;
             }
         }
-        
+
         Ok(())
     }
 }
@@ -343,7 +343,7 @@ impl Writeable for String {
     fn write(&self, buffer: &mut Vec<u8>, version: ProtocolVersion) -> anyhow::Result<()> {
         VarInt(self.len() as i32).write(buffer, version)?;
         buffer.extend_from_slice(self.as_bytes());
-        
+
         Ok(())
     }
 }
@@ -417,7 +417,9 @@ where
 {
     fn write(&self, buffer: &mut Vec<u8>, version: ProtocolVersion) -> anyhow::Result<()> {
         P::try_from(self.0.len())?.write(buffer, version)?;
-        self.0.iter().for_each(|item| item.write(buffer, version).expect("failed to write to vec"));
+        self.0
+            .iter()
+            .for_each(|item| item.write(buffer, version).expect("failed to write to vec"));
 
         Ok(())
     }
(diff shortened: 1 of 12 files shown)


feather/protocol/src/io.rs
```diff
@@ -388,51 +392,57 @@ where
...
+impl<'a, P, T> Writeable for LengthPrefixedVec<'a, P, T>
 where
     T: Writeable,
     [T]: ToOwned<Owned = Vec<T>>,
+    P: TryInto<usize> + TryFrom<usize> + Writeable,
+    <P as TryFrom<usize>>::Error: std::fmt::Debug,
 {
     fn write(&self, buffer: &mut Vec<u8>, version: ProtocolVersion) {
-        VarInt::from(self.0.len()).write(buffer, version);
+        P::try_from(self.0.len()).unwrap().write(buffer, version);
```
kirawi, line 407: I didn't know how to handle this error, so I just unwrapped it. Will it cause problems?

feather/protocol/src/io.rs
```diff
@@ -388,51 +392,57 @@ where
...
+        Ok(Self(Cow::Owned(vec), PhantomData))
     }
 }
 
-impl<'a, T> Writeable for LengthPrefixedVec<'a, T>
+impl<'a, P, T> Writeable for LengthPrefixedVec<'a, P, T>
 where
     T: Writeable,
     [T]: ToOwned<Owned = Vec<T>>,
+    P: TryInto<usize> + TryFrom<usize> + Writeable,
```
Schuwi, line 403: Is there a reason to have the `TryInto` bound?

feather/protocol/src/io.rs
Schuwi: Hmm yeah not sure right now. Panicking should be avoided whenever necessary and reaching 65000 vec elements in case of a ShortPrefixedVec is not completely unrealistic.
Not sure what the best course of action would be

Schuwi:
Looks very good overall! :)

feather/protocol/src/io.rs
kirawi: I'm considering making the function return a result, but I don't really know if the error can be propagated in all instances.

kirawi:
I made the trait return a result, and all other functions simply unwrap it. I'll probably change it to have proper error handling in the future.

kirawi:
Okay, I think it should be good enough for now. Ready for review.

feather/protocol/src/io.rs
```diff
@@ -388,51 +404,57 @@ where
...
+    P: TryFrom<usize> + Writeable,
+    P::Error: std::error::Error + Send + Sync + 'static,
 {
-    fn write(&self, buffer: &mut Vec<u8>, version: ProtocolVersion) {
-        VarInt::from(self.0.len()).write(buffer, version);
+    fn write(&self, buffer: &mut Vec<u8>, version: ProtocolVersion) -> anyhow::Result<()> {
+        P::try_from(self.0.len())?.write(buffer, version)?;
+        self.0
+            .iter()
+            .for_each(|item| item.write(buffer, version).expect("failed to write to vec"));
```
kirawi, line 422: Is there a better way to handle this error? I'd like to propagate it somehow.

caelunshun (approved):
Thanks!

PR #415, opened 2021-04-18
WorldSource for the Worldgen

michidk:
# WorldSource for the Worldgen

## Status

- [x] Ready 

## Description

I implemented a WorldSource for the worldgen

## Related issues

_Leave empty if none_

## Checklist

### DIFF
--- a/feather/common/src/world_source/world_gen.rs
+++ b/feather/common/src/world_source/world_gen.rs
@@ -1,0 +1,95 @@
+use super::{ChunkLoadResult, LoadedChunk, WorldSource};
+use base::ChunkPosition;
+use feather_worldgen::{ComposableGenerator, WorldGenerator};
+use flume::{Receiver, Sender};
+use std::time::Duration;
+
+// World source that uses the world generator crate
+pub struct WorldGenWorldSource {
+    request_sender: Sender<ChunkPosition>,
+    result_receiver: Receiver<LoadedChunk>,
+}
+
+impl WorldGenWorldSource {
+    pub fn new() -> Self {
+        let gen = ComposableGenerator::default_with_seed(0);
+
+        let (request_sender, request_receiver) = flume::unbounded();
+        let (worker, result_receiver) = Worker::new(request_receiver, gen);
+
+        worker.start();
+
+        Self {
+            request_sender,
+            result_receiver,
+        }
+    }
+}
+
+impl WorldSource for WorldGenWorldSource {
+    fn queue_load(&mut self, pos: ChunkPosition) {
+        self.request_sender
+            .send(pos)
+            .expect("chunk worker panicked");
+    }
+
+    fn poll_loaded_chunk(&mut self) -> Option<super::LoadedChunk> {
+        self.result_receiver.try_recv().ok()
+    }
+}
+
+impl Default for WorldGenWorldSource {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+struct Worker {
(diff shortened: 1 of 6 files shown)


feather/common/src/world_source/world_gen.rs
```diff
@@ -0,0 +1,95 @@
+use super::{ChunkLoadResult, LoadedChunk, WorldSource};
+use base::ChunkPosition;
+use feather_worldgen::{ComposableGenerator, WorldGenerator};
+use flume::{Receiver, Sender};
+use std::time::Duration;
+
+// World source that uses the world generator crate
```
Schuwi, line 7: This should be a doc comment (`///`)

feather/worldgen/src/lib.rs
```diff
@@ -141,8 +141,8 @@ impl WorldGenerator for ComposableGenerator {
 
         let mut chunk = Chunk::new(position);
 
-        for x in 0..16 {
-            for z in 0..16 {
+        for x in 0..4 {
+            for z in 0..4 {
```
Schuwi, line 145: This... doesn't seem right but I don't claim to 100% understand what's going on. Why is this change necessary?

feather/worldgen/src/finishers/clumped.rs
```diff
@@ -41,7 +41,7 @@ impl FinishingGenerator for ClumpedFoliageFinisher {
                             let pos_x = cmp::max(0, cmp::min(x as i32 + offset_x, 15)) as usize;
                             let pos_z = cmp::max(0, cmp::min(z as i32 + offset_z, 15)) as usize;
 
-                            if chunk.biomes().get(pos_x, 0, pos_z) != biome {
+                            if chunk.biomes().get(pos_x / 4, 0, pos_z / 4) != biome {
```
Schuwi, line 44: Same here, why is this change necessary?

Schuwi:
The problem seems to be that with 1.16 Minecraft switched to using a 3D grid for biomes and simultaneously changed the resultion to 4x4x4 biome cells per chunk section instead of the 16x16 (biome per block column) resolution as before.
See here: https://github.com/feather-rs/feather/blob/679bc6e8c988425ef448004ed806120190c62164/feather/base/src/chunk/biome_store.rs#L11-L20

While `feather-base` is updated to the new system, `feather-worldgen` doesn't seem to be updated yet:
https://github.com/feather-rs/feather/blob/0f6f217af81d753d7b697b80a8de206eb9cf5407/feather/worldgen/src/lib.rs#L324-L330

Schuwi (changes requested):
The problem seems to be that with 1.16 Minecraft switched to using a 3D grid for biomes and simultaneously changed the resultion to 4x4x4 biome cells per chunk section instead of the 16x16 (biome per block column) resolution as before.
See here: https://github.com/feather-rs/feather/blob/679bc6e8c988425ef448004ed806120190c62164/feather/base/src/chunk/biome_store.rs#L11-L20

While `feather-base` is updated to the new system, `feather-worldgen` doesn't seem to be updated yet:
https://github.com/feather-rs/feather/blob/0f6f217af81d753d7b697b80a8de206eb9cf5407/feather/worldgen/src/lib.rs#L324-L330

michidk:
Yeah, you are right. I thought of my changes as more of an improvement (the code that is broken is already in this repo, - now at least you can try out the world generation (without biomes)) than a rewrite for the whole generation system (which Iam not too familiar with either). Closing this for now.

Issue #422, opened 2021-05-11
Feature: PlayerAbilities


Miro-Andrin:
# Feature Request
Handle PlayerAbilities request in [feather/server/src/packet_handlers.rs](https://github.com/feather-rs/feather/blob/b63fd95da0170eaa89f7241f0749bea0987262f8/feather/server/src/packet_handlers.rs#L69)

## Description
[wiki](https://wiki.vg/Protocol#Player_Abilities_.28serverbound.29)
The vanilla client sends this packet when the player starts/stops flying (like in creative mode).

## Describe the solution you'd like
Add a component IsCreativeFlying on Player entities that track if a player is flying or not, and add a event CreativeFlyingChange. When this packet is recived we update the IsCreativeFlying component and issue the CreativeFlyingChange event. 
Somewhere we also need to keep track if the player needs to be kicked if they are not in creative mode, but that could reasonably be handled elsewhere, or left out for plugins to handle. 

#### 1)
Create a IsCreativeFlying component in [quill](https://github.com/feather-rs/feather/blob/b63fd95da0170eaa89f7241f0749bea0987262f8/quill/common/src/components.rs). 

#### 2)
Create a CreativeFlyingChange Event. This is almost the same steps as for 1), but you should look at [quill/common/events](https://github.com/feather-rs/feather/tree/main/quill/common/src/events)

#### 3)
Add a handle function in [feather/server/src/packet_handlers.rs](https://github.com/feather-rs/feather/blob/b63fd95da0170eaa89f7241f0749bea0987262f8/feather/server/src/packet_handlers.rs#L69) for handling the PlayerAbilities packet, that issues the event and updates the component.  For an example on how to trigger the event go to [feather/server/src/packet_handlers/interaction.rs](https://github.com/feather-rs/feather/blob/b63fd95da0170eaa89f7241f0749bea0987262f8/feather/server/src/packet_handlers/interaction.rs#L89), or go directly to the [method](https://github.com/feather-rs/feather/blob/b63fd95da0170eaa89f7241f0749bea0987262f8/feather/ecs/src/event.rs#L39). 

#### 4)
Create a is_creative_flying_check system in [feather/server/src/systems](https://github.com/feather-rs/feather/tree/main/feather/server/src/systems) that listens for the CreativeFlyingChange event and kicks players if they are not in creative mode (and they are turning flying on). 

[...]

PR #425, opened 2021-05-11
Player abilities packet

Miro-Andrin:
Implemented everything in #422,  except the parts about kicking players that use this packet when not in creative mode. I also added a small example plugin in quill that prints a short msg when a player starts or stops flying. 

I have run:
```
cargo clippy --all -- -D warnings
```
And eventually:
```
cargo fmt
```
And tested the example plugin.


caelunshun (approved):
Thank you!

PR #430, opened 2021-06-03
[WIP] Implement Potion Effects

walker27460:
# Implement Potion Effects

## Status

- [ ] Ready 
- [x] Development
- [ ] Hold

## Description

Implemented Potion Effects.

## Checklist

### DIFF
--- a/feather/common/src/entities/player.rs
+++ b/feather/common/src/entities/player.rs
@@ -1,13 +1,14 @@
 use anyhow::bail;
 use base::EntityKind;
 use ecs::{EntityBuilder, SysResult};
-use quill_common::{components::CreativeFlying, entities::Player};
+use quill_common::{components::{CreativeFlying, PotionEffects}, entities::Player};
 
 pub fn build_default(builder: &mut EntityBuilder) {
     super::build_default(builder);
     builder
         .add(Player)
         .add(CreativeFlying(false))
+        .add(PotionEffects(vec![]))
         .add(EntityKind::Player);
 }
 
--- a/feather/common/src/entities/player.rs
+++ b/feather/common/src/entities/player.rs
@@ -11,7 +11,6 @@ pub fn build_default(builder: &mut EntityBuilder) {
     builder
         .add(Player)
         .add(CreativeFlying(false))
-        .add(PotionEffects(vec![]))
         .add(Sneaking(false))
         .add(EntityKind::Player);
 }
--- a/feather/common/src/entities/player.rs
+++ b/feather/common/src/entities/player.rs
@@ -2,7 +2,7 @@ use anyhow::bail;
 use base::EntityKind;
 use ecs::{EntityBuilder, SysResult};
 use quill_common::{
-    components::{CreativeFlying, PotionEffects, Sneaking},
+    components::{CreativeFlying, Sneaking},
     entities::Player,
 };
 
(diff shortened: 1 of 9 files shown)


feather/common/src/entities/player.rs
```diff
@@ -11,6 +11,7 @@ pub fn build_default(builder: &mut EntityBuilder) {
     builder
         .add(Player)
         .add(CreativeFlying(false))
+        .add(PotionEffects(vec![]))
```
OxKitsune, line 14: Do we need this? Is it not better to only add the component when the player actually has potion effects?

OxKitsune:
Maybe some utility methods for adding potion effects to players need to be added? Just like the particle spawning methods, cause you could just add the components to the player. But I think having a util method would be good.

feather/common/src/entities/player.rs
walker27460: maybe that will be better?
if so then I'm gonna remove that line

quill/common/src/components.rs
```diff
@@ -106,20 +107,73 @@ pub struct CreativeFlying(pub bool);
...
+impl_effect!(Saturation);
+impl_effect!(Glowing);
+impl_effect!(Levitation);
+impl_effect!(Luck);
+impl_effect!(BadLuck);
+impl_effect!(SlowFalling);
+impl_effect!(ConduitPower);
+impl_effect!(DolphinsGrace);
+impl_effect!(BadOmen);
+impl_effect!(HeroOfTheVillage);
```
amberkowalski, line 176: All of these should have the `Effect` suffix.

quill/common/src/components.rs
```diff
@@ -106,20 +107,73 @@ pub struct CreativeFlying(pub bool);
...
-    particle: bool,
-    ambient: bool, // given from beacon or not.
-    icon: bool,    // show effect icon.
+#[derive(Copy, Clone, Debug, Hash, Serialize, PartialEq, Eq, Deserialize)]
+pub struct PotionApplication {
+    pub amplifier: u8,
+    pub duration: u32,
+    pub particle: bool,
+    pub ambient: bool, // given from beacon or not.
+    pub icon: bool,    // show effect icon.
```
amberkowalski, line 116: Switch the comments to doc comments on the members, consider adding one for `particle` and `duration`.

quill/common/src/components.rs
```diff
@@ -106,20 +107,73 @@ pub struct CreativeFlying(pub bool);
...
+}
+
+macro_rules! impl_effect {
+    ($ident:ident) => {
+        #[derive(Serialize, Deserialize, Eq, PartialEq, Hash)]
+        pub struct $ident(BTreeSet<PotionApplication>);
+        impl $ident {}
+        bincode_component_impl!($ident);
+    };
+}
```
amberkowalski, line 143: I think we might have somewhere we try to consolidate macros? Not sure.

quill/common/src/components.rs
amberkowalski, line 116: We could probably move effects into their own module.

amberkowalski (commented):
Just a few things I see, I know its not done yet.

quill/common/src/components.rs
walker27460: the effect enum was originally in libcraft but I made a transition to components.

amberkowalski:
Hi there, it appears your PR has been inactive for a while. Are you planning to continue it? Please let me know so I can either pick it up or wait for more progress.

walker27460:
yeah im try to continue but no progress

Defman:
Bump

PR #433, opened 2021-06-07
Add the ability to create queries that mutate host components within plugins

ambeeeeee:
# Query with mutable borrows

## Status

- [X] Ready 
- [ ] Development
- [ ] Hold

## Description

Adds the ability to perform `for (entity, (mut position, cow)) in game.query::<(&mut Position, &Cow)>` within plugins.

### Implementation
 
 This PR adds a `Query` impl for `&'a mut T` alongside the existing implementation for `&'a T`.   

Additionally, it adds `borrowed_mut` to `Query` to indicate if a `HostComponent` is borrowed mutably within the `Query`.   
This is combined in the tuple query, and then at iterator creation it counts each instance of `HostComponent` within the `Query`'s list of components and ensures if one is found more than once that that component is not borrowed mutably. 

### Notable Changes

Quill's `Entity::set` has been changed to `Entity::insert` because its primary purpose is no longer updating a component on an entity.

## Notes for the future

- We should examine the performance effects and correctness impacts of queuing up query mutations and submit them to the host in bulk.
- We may want to support nesting queries in the future, this solution will not comply with that. This can be trivially worked around for now, and we can examine supporting nesting in the future. This will require a more elaborate borrow checker, which is why I'm avoiding it for now.

## Note for reviewers

Is this solution 100% perfect and awesome? No. There are definite improvements we could make sometime in the future but for now this is an entirely reasonable step in that direction. Any future changes won't cause backwards incompatibility, so its safe to put them off for the future. Its important that we get feather moving forward so we shouldn't seek absolute perfection for this feature.

## Checklist

- [X] Ran `cargo fmt`, `cargo clippy`, `cargo build --release` and `cargo test` and fixed any generated errors!
- [X] Removed unnecessary commented out code
- [ ] ~~Used specific traces (if you trace actions please specify the cause i.e. the player)~~

### DIFF
--- a/quill/api/Cargo.toml
+++ b/quill/api/Cargo.toml
@@ -15,4 +15,5 @@ quill-sys = { path = "../sys" }
 quill-common = { path = "../common" }
 thiserror = "1"
 uuid = "0.8"
+itertools = "0.10.0"
 
--- a/quill/api/src/game.rs
+++ b/quill/api/src/game.rs
@@ -109,7 +109,7 @@ impl Game {
     ///    println!("Found an entity with position {:?} and UUID {}", position, uuid);
     /// }
     /// ```
-    pub fn query<Q: Query>(&self) -> QueryIter<Q> {
+    pub fn query<Q: Query>(&mut self) -> QueryIter<Q> {
         QueryIter::new()
     }
 
--- a/quill/api/src/query.rs
+++ b/quill/api/src/query.rs
@@ -91,11 +91,7 @@ where
     }
 
     fn borrowed_mut(ty: HostComponent) -> bool {
-        if ty == T::host_component() {
-            true
-        } else {
-            false
-        }
+        ty == T::host_component()
     }
 
     unsafe fn get_unchecked(
@@ -182,13 +178,11 @@ where
         Q::add_component_types(&mut component_types);
 
         for (component_type, count) in component_types.clone().into_iter().counts() {
-            if count > 1 {
-                if Q::borrowed_mut(component_type) {
-                    panic!(
-                        "{:?} was borrowed mutably and immutably at the same time",
-                        component_type
-                    )
-                }
+            if count > 1 && Q::borrowed_mut(component_type) {
+                panic!(
+                    "{:?} was borrowed mutably and immutably at the same time",
+                    component_type
+                )
             }
         }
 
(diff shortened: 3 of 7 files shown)


quill/api/Cargo.toml
```diff
@@ -15,4 +15,5 @@ quill-sys = { path = "../sys" }
 quill-common = { path = "../common" }
 thiserror = "1"
 uuid = "0.8"
+itertools = "0.10.0"
```
Defman, line 18: I hate itertools... maybe its just me.

quill/api/src/game.rs
```diff
@@ -109,7 +109,7 @@ impl Game {
     ///    println!("Found an entity with position {:?} and UUID {}", position, uuid);
     /// }
     /// ```
-    pub fn query<Q: Query>(&self) -> QueryIter<Q> {
+    pub fn query<Q: Query>(&mut self) -> QueryIter<Q> {
```
Defman, line 112: Hmm I don't think quires should mutable borrow game, since you can write read only queries. It should be checked at runtime.

quill/api/src/query.rs
```diff
@@ -1,19 +1,28 @@
...
 /// A type that can be used for a query.
 ///
 /// Implemented for tuples of `Query`s as well.
 pub trait Query {
     type Item;
+    type Target;
 
     fn add_component_types(types: &mut Vec<HostComponent>);
 
+    fn borrowed_mut(ty: HostComponent) -> bool;
```
Defman, line 24: This can be simplified to
```rust
const MUT: bool = true;
const MUT: bool = false;
```

Defman (changes requested):
Hmm I'm not the best to check this, but it looks "good" to my eyes.

quill/api/src/game.rs
amberkowalski: I think for now it should be fine? *Theoretically* the new ECS will come eventually and fix this.

quill/api/src/query.rs
amberkowalski: I don't think this can be, see the tuple query impl

Defman: Ahh so you recursively call borrow_mut down, in the tuple impl... you can do that with type system only.

amberkowalski: Let see that then?

amberkowalski: I don't wannt toooooo I just wanna merge thissss 😭

Defman:
Ready to be merged?

ambeeeeee:
Sure

Defman (approved):

PR #439, opened 2021-06-27
Made it so quill::Plugin does not have to be in scope for plugins.

Miro-Andrin:
# Small fix for plugins.

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Description
The following code used to not compile, because quill::Plugin was not in scope. Now it does.
```rust
pub struct CommandPlugin {}

quill::plugin!(CommandPlugin);

impl quill::Plugin  for CommandPlugin {
    fn enable(game: &mut quill::Game, setup: &mut quill::Setup<Self>) -> Self {
        CommandPlugin{}
    }

    fn disable(self, game: &mut quill::Game) {
        todo!()
    }
}
```
I modified the quill::plugin! macro to not rely on quill::Plugin to be in scope. 

## Checklist

### DIFF
--- a/quill/api/src/lib.rs
+++ b/quill/api/src/lib.rs
@@ -94,10 +94,11 @@ macro_rules! plugin {
         #[doc(hidden)]
         #[cfg(target_arch = "wasm32")]
         pub unsafe extern "C" fn quill_setup() {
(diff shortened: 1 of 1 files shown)


caelunshun (approved):
Thanks!

The CI failure is caused by Clippy changes in 1.53 unrelated to this PR, so I'll merge.

Issue #435, opened 2021-06-11
Missing "pull_request_template.md" under /docs/


WashingtonARamos:
[CONTRIBUTING.md](https://github.com/feather-rs/feather/blob/main/CONTRIBUTING.md#notes-to-your-code) says:

>## Notes to your code
>For notes to your code check the Checklist from [`pull_request_template.md`](docs/pull_request_template.md)

But said file is missing.

OxKitsune:
Its located at  `.github/pull_request_template.md`

FussballAndy:
Oh I moved it to .github/pull_request_template.md forgot there was a reference in CONTRIBUTING.md.

Issue #418, opened 2021-05-05
Server crashed after spamming it with many clients


Seadragon91:
## Description
Spammed the server with many clients using pycraft. The clients joined the server, disconnected and connected again in a loop.

## Reproduction Steps

1. Spam the server with many clients and disconnect and reconnect them very quickly.

## What You Expected to Happen
Server should be able to handle it, when many clients connects and disconnects.

## What Actually Happened
Server crashes after spamming it with many clients.

## Screenshots and Logs
```thread main' panicked at 'called `Option::unwrap()` on a `None` value', feather/server/src/systems/view.rs:43:52```
https://github.com/feather-rs/feather/blob/b63fd95da0170eaa89f7241f0749bea0987262f8/feather/server/src/systems/view.rs#L43
```
   0: rust_begin_unwind
             at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/core/src/panicking.rs:50:5
   3: core::option::Option<T>::unwrap
             at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/core/src/option.rs:386:21
   4: feather_server::systems::view::send_new_chunks
             at /home/lukas/rust/feather/feather/server/src/systems/view.rs:43:22
   5: core::ops::function::FnMut::call_mut
             at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/core/src/ops/function.rs:150:5
   6: feather_ecs::system::GroupBuilder<Input,State>::make_function::{{closure}}
             at /home/lukas/rust/feather/feather/ecs/src/system.rs:193:13
   7: <alloc::boxed::Box<F,A> as core::ops::function::FnMut<Args>>::call_mut
             at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/alloc/src/boxed.rs:1335:9
   8: feather_ecs::system::SystemExecutor<Input>::run
             at /home/lukas/rust/feather/feather/ecs/src/system.rs:144:26
   9: feather_server::create_tick_loop::{{closure}}
[...]

Issue #411, opened 2021-04-16
Player Spawns Twice


zNoctum:
## Description
I connect twice to the server with the *same* Client. and it creates two Players

## Reproduction Steps

1. Start latest from source build with +nightly and not release Opt
2. start the server
3. connect while start up twice in rapid succession
4. You will now be connect the second you will spawn in a few seconds

## What You Expected to Happen
Only one Me should spawn

## Screenshots and Logs

![double-player](https://user-images.githubusercontent.com/46860055/115065920-157fc700-9eef-11eb-937a-2a67c3af638f.png)

## Your Environment
- Operating System: Arch linux
- **Feather** Version: latest

PR #440, opened 2021-06-28
Bug fixes

FussballAndy:
# Bug fixes

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Description

- Fixed it that one profile could login multiple times
- Fixed the server crashing when spammed with clients/Made it so that chunk updates check if the player is still there
- Fixed Link to `pull_request_template.md` in `CONTRIBUTING.md`
- Fixed clippy errors for 1.53

## Related issues
#411 #418 #435 

## Checklist

### DIFF
--- a/feather/server/src/lib.rs
+++ b/feather/server/src/lib.rs
@@ -98,11 +98,7 @@ impl Server {
     pub fn accept_new_players(&mut self) -> Vec<ClientId> {
         let mut clients = Vec::new();
         for player in self.new_players.clone().try_iter() {
-            if let Some(cl) = self
-                .clients
-                .iter()
-                .find(|x| *x.username() == player.username)
-            {
+            if let Some(cl) = self.clients.iter().find(|x| x.uuid() == player.uuid) {
                 cl.disconnect("Logged in from another location!");
             }
             let id = self.create_client(player);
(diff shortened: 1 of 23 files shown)


feather/server/src/lib.rs
```diff
@@ -98,6 +98,9 @@ impl Server {
     pub fn accept_new_players(&mut self) -> Vec<ClientId> {
         let mut clients = Vec::new();
         for player in self.new_players.clone().try_iter() {
+            if let Some(cl) = self.clients.iter().find(|x| x.uuid() == player.uuid) {
+                cl.disconnect("Logged in from another location!");
```
caelunshun, line 102: Looks like we need to `return` here, or else the new client gets added anyway.

feather/server/src/systems/view.rs
```diff
@@ -40,16 +40,20 @@ fn send_new_chunks(game: &mut Game, server: &mut Server) -> SysResult {
...
-        client.update_own_chunk(event.new_view.center());
-        update_chunks(
-            game,
-            player,
-            client,
-            event,
-            position,
-            &mut server.waiting_chunks,
-        )?;
+        // As each iteration takes a while, it can happen that a player
```
caelunshun, line 43: The cause of the problem isn't actually clients being removed while the loop is running, since the client list is only accessible by one thread. It happens because players in the ECS aren't removed for one extra tick after they disconnect (so plugins can handle disconnect events), and as a result, the query finds dead clients.

The solution here still works, but the comment should be changed.

feather/server/src/lib.rs
FussballAndy: We are not removing the new client but instead the old one. Thus returning would probably just break everything.

caelunshun (approved):
Thank you!

PR #442, opened 2021-08-01
Worldgen2

koskja:
# Reimplementing world generation

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Description

Added a new WorldSource that generates new chunks, updated chunk biomes for 1.16, redid flat world gen and tweaked default worldgen to a more normal state.

## Related issues

World generation #24 

## Checklist


feather/common/src/world_source/generating.rs
```diff
@@ -0,0 +1,67 @@
...
+
+impl GeneratingWorldSource {
+    pub fn from_arc<G>(generator: Arc<G>) -> Self
+    where
+        G: WorldGenerator + 'static,
+    {
+        let (send_pos, recv_pos) = unbounded::<ChunkPosition>();
+        let (send_gen, recv_gen) = unbounded::<LoadedChunk>();
+        for _ in 0..8 {
+            // FIXME: do not hardcode '8' as a magic number
```
caelunshun, line 28: I'd recommend starting just one thread here that polls the channel, then invokes `rayon::spawn` for each chunk that needs to be generated. Right now, we're blocking the Rayon thread poll by waiting on channels, which can cause a deadlock

feather/server/config.toml
```diff
@@ -12,7 +12,7 @@ online_mode = true
 motd = "A Feather server"
 max_players = 16
 default_gamemode = "creative"
-view_distance = 12
+view_distance = 32
```
caelunshun, line 15: This should probably be reverted? 12 is the default for vanilla too.

koskja: I thought that the global threadpool can have an unlimited amount of threads? Also, while the channels are blocked, other threads can work in the meantime. And of course, the view_distance should be set to 12.

caelunshun: > I thought that the global threadpool can have an unlimited amount of threads?

It's a fixed size AFAIK. See https://docs.rs/rayon-core/1.9.1/rayon_core/struct.ThreadPoolBuilder.html#method.num_threads

> Also, while the channels are blocked, other threads can work in the meantime.

Yes, but the Rayon threads are blocked. Which means other tasks on the Rayon thread pool can't run.

koskja: Just learned that rayon threads are not the same as os threads :) I am not a fan of making a separate thread for every chunk, do you think a 'personal' ThreadPool would be a good idea?

caelunshun: Using `rayon::spawn` won't create a new thread for each chunk. It dispatches a task into a pool of existing threads, which is very cheap.

koskja:
review?

caelunshun (approved):
LGTM. Thank you!

Issue #429, opened 2021-06-03
Features: Varius Registries Needed For Loading Datapacks


Miro-Andrin:
# Feature Request
To implement loading data from [datapacks](https://minecraft.fandom.com/wiki/Data_Pack#Contents)  we need to load various different values into feather and store them somehow. Although the [ecs](https://en.wikipedia.org/wiki/Entity_component_system) is useful for some of those, we however probably don't want to use it for everything. We therefore want to implement some registries that store things more efficiently and provide better ergonomics. 

I am going to start by listing all the values stored in data packs. I think all of them should be stored in registries, but if you disagree then comment, because we can't have someone being wrong on the internet.  

- [Advancements](https://minecraft.fandom.com/wiki/Advancement/JSON_format)
- [Dimensions](https://minecraft.fandom.com/wiki/Custom_dimension)
- [Functions](https://minecraft.fandom.com/wiki/Function_(Java_Edition))
- [Loot Tables](https://minecraft.fandom.com/wiki/Loot_table)
- [Predicates](https://minecraft.fandom.com/wiki/Predicate)
- [Reipies](https://minecraft.fandom.com/wiki/Recipe)
- [Structures](https://minecraft.fandom.com/wiki/Structure_Block_file_format)
- [Tags](https://minecraft.fandom.com/wiki/Tag)
- [World Generation Settings](https://minecraft.fandom.com/wiki/Custom_world_generation)
- [Biomes](https://minecraft.fandom.com/wiki/Biomes)

Together with these registries we also need to implement host_calls (see feather/plugin-host/src/host_calls.rs and  /feather/quill/sys/src/lib.rs)  such that we can edit them from plugins. Any change in these registries should probably also trigger change events. However i don't know what entity that event  would be connected to?  If you want a pointer on how to implement a event, then look at how i implemented [#425](https://github.com/feather-rs/feather/pull/425)

## Some implementation questions.
Should we store all these registries together in one directory? I suggest '/feather/common/registries', does that seem fine?  Do we need to implement serialisation/deserialisation such that their values can be saved once edited by a plugin? 

## Wrapping up: 
If you are looking for something to do then i suggest working on Advancements, Functions, Loot Tables, or Recipes because they seem more immediately necessary.  @Defman  has started on recipes, but if someone is quick enough they can finish it before him :1st_place_medal:

And feel free to ask in the discord if you have any questions .

PR #455, opened 2021-08-24
Datapack parsing

koskja:
# Datapack parsing

## Status

- [ ] Ready 
- [x] Development
- [ ] Hold

## Description

The game needs to parse registries for most of its logic to function. This pull request attempts to do that.

## Related issues

#429

## Checklist


koskja:
The functions now take `T: Into<NamespacedId> + Clone`. @caelunshun

feather/datapacks/tags.py
```diff
@@ -0,0 +1,42 @@
+from os import listdir
```
Defman, line 1: Should be moved into the generator directory.

Defman (changes requested):
Looks good, the location of the generator.py is a bit odd and should be moved into the same directory with all other generators?

koskja:
Moved the tags generator. I believe the recipe implementation can be finished together with crafting.

Issue #177, opened 2020-03-02
Block placement / interaction


Defman:
The `PlayerBlockPlacement` packet is also responsible for any block interaction ie. opening chests or filling bottles from a cauldron. The current block placement system have to be reworked into the following three steps.

1. Try to interact with the block
2. Try to place the held item at the location
3. Try to place the held item at the location + face

1. Interacting with blocks simple match over held item and block.
```rust
match (held_item, from_block) {
  (_, Block::Chest(_)) => open_chest(),
  (Item::Bottle, Block::Cauldron(_)) => fill_bottle(),
  ...
}
```
2. Simple match on the from block, ie. can the block be replaced
```rust
match from_block {
  Block::Air | Block::CaveAir | Block::Grass ... => place_block
}
// Handle waterlog
match (held_item, from_block) {
   (Item::WaterBucket, Block::Fence) => Block::Fence(data.waterlogged = true),
  ...
}
```
3. This step is identical to step 2. but the from block = state.block_at(location + face).

caelunshun:
It might be wise to use a different approach than the `match` for interactions—as the number of blocks supporting interactions increases, this function is going to get pretty unwieldy. Would something like the [inventory](https://docs.rs/inventory/) crate be of interest here?

Defman:
It seems like [inventory](https://docs.rs/inventory/0.1.5/inventory/) would be a nice fit, however I had thought about using the event system. Such that an interaction event is first fired and it has a mutable `canceled` field which any event handler must set to true if the interaction was successful. Second a block place event is fired at the location if the previous event was unsuccessful and last a block place event at `location + face`.

I dont know how to add a mutable field to an event, if it is even possible without a mutex guard. If so it would be a synchronization overhead which would negatively effect performance.

Defman:
Seems like tonks already uses inventory for event handler registration?

AnEthernetCable:
It appears block placement does not have any checks to make sure it is not blocked by an entity (ex: players) - I don't believe this is a known issue at the moment

PR #456, opened 2021-09-01
Block placement

koskja:
# Block placement

## Status

- [ ] Ready 
- [x] Development
- [ ] Hold

## Description

Working on placing blocks so that they persist between loads. Other players also see this.

## Related issues

#177 

## Checklist

### DIFF
--- a/feather/common/src/block.rs
+++ b/feather/common/src/block.rs
@@ -1,0 +1,77 @@
+use base::{Area, BlockId, Gamemode, Inventory, Item, ItemStack};
+use ecs::{SysResult, SystemExecutor};
+use quill_common::events::BlockPlacementEvent;
+use crate::chunk::entities::ChunkEntities;
+use crate::entities::player::HotbarSlot;
+use crate::events::BlockChangeEvent;
+use crate::{Game, World};
+
+pub fn register(systems: &mut SystemExecutor<Game>) {
+    systems.add_system(block_placement);
+}
+
+pub fn block_placement(game: &mut Game) -> SysResult {
+    let mut events = vec![];
+    for (player, event) in game.ecs.query::<&BlockPlacementEvent>().iter() {
+        let inv = game.ecs.get_mut::<Inventory>(player)?;
+        let gamemode = game.ecs.get::<Gamemode>(player)?;
+        let hotbar = game.ecs.get::<HotbarSlot>(player)?;
+        let mut slot = match event.hand {
+            libcraft_core::Hand::Main => inv.item(Area::Hotbar, hotbar.get()),
+            libcraft_core::Hand::Offhand => inv.item(Area::Offhand, 0),
+        }.unwrap();
+        if slot.is_none() { continue; }
+        let block = match item_to_block(slot.as_ref().unwrap().item()) {
+            Some(s) => s,
+            None => continue,
+        };
+        match place_block(&mut game.world, &game.chunk_entities, block, event) {
+            Some(s) =>  {
+                match *gamemode {
+                    Gamemode::Survival | Gamemode::Adventure => decrease_slot(&mut slot),
+                    Gamemode::Creative | Gamemode::Spectator => {},
+                }
+                events.push(s)
+            },
+            None => {},
+        }
+    }
+    for e in events {
+        game.ecs.insert_event(e);
+    }
+    
+    Ok(())
+}
+
+fn place_block(world: &mut World, chunk_entities: &ChunkEntities, block: BlockId, placement: &BlockPlacementEvent) -> Option<BlockChangeEvent> {
+    let placement_pos = placement.location.adjacent(placement.face);
+    if !world.block_at(placement_pos).map(|e| e.is_replaceable()).unwrap_or(false) { // Not loaded or solid
+        return None;
+    }
+    if chunk_entities.entities_in_chunk(placement_pos.chunk()).iter().any(|_entity| { false }) { // FIXME: Somehow check if block would collide with any entities
+        return None;
+    }
+    if world.set_block_at(placement_pos, block) {
+        Some(BlockChangeEvent::single(placement_pos))
+    } else {
+        None
+    }
+}
+
+fn item_to_block(item: Item) -> Option<BlockId> {
+    let mut name = "minecraft:".to_owned();
+    name.push_str(item.name());
+    BlockId::from_identifier(&name)
+}
+
+fn decrease_slot(slot: &mut Option<ItemStack>) {
+    match slot  {
+        Some(s) => {
+            s.remove(1);
+            if s.count() == 0 {
+                *slot = None;
+            }
+        },
+        None => {},
+    }
+}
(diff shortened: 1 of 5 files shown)


koskja:
Almost done, some things that are still missing are proper placement checking (e. g. you cannot place sugarcane on dry land) and fence/wall connecting. Signs, banners, chests etc. don't work properly because block entities aren't implemented.
Somebody could take a look at this.

qualterz:
Noticed a bug. When I place the blocks of bed in a row too quickly, server looks like does not handle some requests from player: [client side view](https://gateway.pinata.cloud/ipfs/Qmf5SoSfPwRNW212D5jKmzs2BiZgh2xRCYyPoAZEBukBD9) and [server side view](https://gateway.pinata.cloud/ipfs/QmfLXfqAeDXWmXL28ibuwJudpNmgb6iHC8zgNWEwLayecJ).

Also, after reconnecting to the server, the beds are not visible: [screenshot](https://gateway.pinata.cloud/ipfs/QmbGQRr691BJniPKBruBhY7HchusbLKQF5fHPaFLFHDFtP).

Defman:
BUMB

koskja:
> Noticed a bug. When I place the blocks of bed in a row too quickly, server looks like does not handle some requests from player: [client side view](https://gateway.pinata.cloud/ipfs/Qmf5SoSfPwRNW212D5jKmzs2BiZgh2xRCYyPoAZEBukBD9) and [server side view](https://gateway.pinata.cloud/ipfs/QmfLXfqAeDXWmXL28ibuwJudpNmgb6iHC8zgNWEwLayecJ).
> 
> Also, after reconnecting to the server, the beds are not visible: [screenshot](https://gateway.pinata.cloud/ipfs/QmbGQRr691BJniPKBruBhY7HchusbLKQF5fHPaFLFHDFtP).

How did you get the serverside view? Not sure how to reproduce on my machine.

Molive-0:
What's stopping this being merged?

Molive-0:
ah

koskja:
Block support checking is missing, but I believe a new PR should be opened for that.

koskja:
Basically done at this point, feel free to test/review.

feather/common/src/block.rs
```diff
@@ -0,0 +1,595 @@
...
+    placement: &BlockPlacementEvent,
+    ecs: &Ecs,
+    light_level: u8,
+) -> Option<Vec<BlockChangeEvent>> {
+    let target1 = placement.location;
+    let target_block1 = world.block_at(target1.try_into().unwrap())?;
+    if target_block1.is_air() {
+        return None;
+    }
+    let mut block = block;
```
Iaiao, line 84: Why not `mut block: BlockId` in a function signature?

feather/common/src/block.rs
```diff
@@ -0,0 +1,595 @@
...
+    let player_dir_ordered = ordered_directions(player_pos.direction());
+    set_face(&mut block, &player_dir_ordered, placement);
+    rotate_8dir(&mut block, player_pos.yaw);
+    // Try placing the block by merging (=placing to the same position as the target block). Includes merging slabs, waterlogging and replacing blocks like grass.
+    let target = if slab_to_place(&mut block, target_block1, placement)
+        | waterlog(&mut block, target_block1)
+        | ((target_block1.kind() != block.kind()) & target_block1.is_replaceable())
+    {
+        merge_slab(&mut block, target_block1);
+        target1
```
Iaiao, line 94: It tries to `merge_slab` if the block is being waterlogged or replaced by another block. `waterlog` already changes the block, so I think you should separate the replacement and merging logic, or at least rename `merge_slab` to something like merge_or_replace. I haven't tested it yet, but I think it's too early to replace the block before stability and entity collision checks. Also, here you're using `|` bitwise operator on booleans, which is incorrect, consider using `||` and `&&` instead of `|` and `&`. iirc there's a clippy lint for this.

feather/common/src/block.rs
```diff
@@ -0,0 +1,595 @@
...
+        None => false,
+    };
+    // This works but there is a discrepancy between when the place block event is fired and getting the entity location.
+    // that makes it possible to place a block at the right exact moment and have the server believe it wasn't blocked.
+    if chunk_entities
+        .entities_in_chunk(target.chunk())
+        .iter()
+        .any(|_entity| {
+            let entity_position = ecs.get::<Position>(*_entity).unwrap();
+            let entity_kind = *ecs.get::<EntityKind>(*_entity).unwrap();
```
Iaiao, line 141: Why is it `_entity` and not just `entity`?

feather/common/src/block.rs
```diff
@@ -0,0 +1,595 @@
...
+    } else if place_head {
+        let face = match block.facing_cardinal()? {
+            FacingCardinal::North => BlockFace::North,
+            FacingCardinal::South => BlockFace::South,
+            FacingCardinal::West => BlockFace::West,
+            FacingCardinal::East => BlockFace::East,
+        };
+        if !world.check_block_stability(block, target.adjacent(face), light_level)? {
+            return None;
+        }
```
Iaiao, line 183: Does the bed really need a stability check? If you just want to make sure it's stable, you should also add this check to place_top (it would be a bit trickier because you'll need to set the lower half of the door, then check and undo it if it's not stable)

feather/common/src/world.rs
```diff
@@ -149,6 +153,248 @@ impl World {
...
+    }
+
+    pub fn get_facing_block(&self, pos: BlockPosition) -> Option<BlockId> {
+        let block = self.block_at(pos.try_into().unwrap())?;
+        let a = block.facing_cardinal().map(FacingCardinal::to_facing_cubic);
+        let b = block
+            .facing_cardinal_and_down()
+            .map(FacingCardinalAndDown::to_facing_cubic);
+        let c = block.facing_cubic();
+        let dir = [a, b, c].iter().find_map(|&e| e)?;
```
Iaiao, line 194: There's `BlockId::has_facing_cardinal()` and other `has_*` functions, so you can do the same thing in an if-else style, and it'll become much more readable.

Iaiao (commented):
I'll test it later.

feather/common/src/block.rs
koskja: True.

feather/common/src/block.rs
koskja: True.

feather/common/src/block.rs
koskja: You are right, it doesn't. This shouldn't a problem with place_top, I don't know of any double block that has extra requirements for its top half.

Iaiao:
There are some issues with connected tripwire (probably also affects rails, didn't test it), door hinge placement on flat surface, and stairs direction when clicking on the side of the block:

https://user-images.githubusercontent.com/55715058/148406567-cc679e49-e04e-4629-bf59-990007e51711.mp4

feather/common/src/world.rs
koskja: The if else chain version is a huge block of text, but I agree that it is easier to understand.

feather/common/src/block.rs
koskja: I will include a little explanation for the logic.

koskja:
Many obvious issues were fixed. What remains is block updates, which are out of scope, and minor details including walls, which are absurdly complicated considering their extremely limited use.

koskja:
I believe that the last thing missing is rails. Considering how bloated this PR is already, this would be possible to implement.
On the other hand, rails have no function without minecarts and their shape greatly depends on block updates.
That is of course up to the devs to decide. So, review? @Defman @Iaiao

PR #458, opened 2021-09-05
Use player.dat

Iaiao:
# player.dat

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Description

This PR makes player's data (inventory, abilities, position, gamemode, previous gamemode, health) persistent by saving it to player.dat files, adds player abilities and previous gamemode component (used in F3+F4 gamemode switcher), methods client.send_abilities, client.set_hotbar_slot, world.load_player_data and world.save_player_data.
This also adds `derive_more` dependency to derive Deref and DerefMut on single-field components.

## Checklist


feather/server/src/systems/player_join.rs
```diff
@@ -26,12 +31,114 @@ fn poll_new_players(game: &mut Game, server: &mut Server) -> SysResult {
...
+        .as_ref()
+        .map(|data| PreviousGamemode::from_id(data.previous_gamemode as i8))
+        .unwrap_or(PreviousGamemode(None));
+
+    client.send_join_game(gamemode, previous_gamemode);
     client.send_brand();
 
-    let mut builder = game.create_entity_builder(Position::default(), EntityInit::Player);
+    // Abilities
+    let walk_speed = player_data
```
koskja, line 49: I think it would be better to move this code to a separate function, in order not to clutter this one.
Maybe `fn player_data_or_default(data: Option<PlayerAbilities>, gamemode: Gamemode) -> PlayerAbilities`, with `Client::send_abilities` taking a `PlayerAbilities` struct instead.

feather/server/src/systems/player_join.rs
```diff
@@ -26,12 +31,114 @@ fn poll_new_players(game: &mut Game, server: &mut Server) -> SysResult {
...
+                z: data.animal.base.position[2],
+                yaw: data.animal.base.rotation[0],
+                pitch: data.animal.base.rotation[1],
+            })
+            .unwrap_or_default(),
+        EntityInit::Player,
+    );
 
     let inventory = Inventory::player();
+    if let Ok(data) = player_data.as_ref() {
```
koskja, line 117: This should also be moved to a separate function.
Very minor nitpick, but I would prefer not silently discarding IO errors.

feather/server/src/systems/player_leave.rs
```diff
@@ -13,12 +21,101 @@ pub fn register(systems: &mut SystemExecutor<Game>) {
...
+    {
         let client = server.clients.get(client_id).unwrap();
         if client.is_disconnected() {
-            server.remove_client(client_id);
             entities_to_remove.push(player);
             broadcast_player_leave(game, name);
+            game.world
+                .save_player_data(
+                    client.uuid(),
+                    &PlayerData {
```
koskja, line 71: This one really asks for a separate function.

koskja (changes requested):
While the code is overall very clean, the `accept_new_player` and `remove_disconnected_clients` should be more concise.

koskja (approved):
Looks good!

ambeeeeee (approved):
Looks okay, not sure why ` PreviousGamemode` is it own type but I assume there's a reason

Iaiao:
> Looks okay, not sure why ` PreviousGamemode` is it own type but I assume there's a reason

Because it's a component, the client uses it in F3+F4 gamemode switcher. It also allows -1 value, which means there's no previous gamemode

Defman (approved):

Defman:
This does not seem to compile on the CI.

Tracreed:
Seems to be the problem of the superflat file adding together str. Some get it and some don't. It hasn't changed so it shouldn't fail.

Defman:
Is it ready to be merged?

Iaiao:
Yes, it's ready

Issue #459, opened 2021-09-07
Moving generated code into libcraft?


Miro-Andrin:
I was looking at adding recipes to libcraft  #429 , but noticed that Item and ItemStack are in feather, under the generated folder.  
Any reason why we should not move all the generated code from feather to libcraft? 

## Description of change.
Move feather/feather/generated/* into feather/libcraft/generated 
And make the generated structs repr-c such that we can use them in quill.

koskja:
I agree, maybe we could also move the feather-generated and the dapapack crate there.

PR #461, opened 2021-09-11
Refactor such that libcraft contains  all generated files (almost).

Miro-Andrin:
# Moved generated code into libcraft. 

## Status
- [x] Ready 
- [ ] Development
- [ ] Hold

## Description
I have moved the code from feather/generated into libcraft/generated and made everything depend on that version instead. I have not moved feather/blocks into libcraft, but might do that in a different pull request.

## Related issues
#459 

## What's missing in this pull request:
ItemStack in libcraft was different in libcraft then in the original generated module. I have no idea why.  The original version allowed for ItemStacks containing  zero items, but the new version does not. This broke a lot of small things, but especially feather/common/window.rs. I tried to quickly adapt it to the new ItemStack methods in libcraft, but i have not gotten it to work. It compiles, but the test don't pass .

## Checklist
- [ x] Ran `cargo fmt`, `cargo clippy --all-targets`, `cargo build --release` and `cargo test` and fixed any generated errors!
- [x] Removed unnecessary commented out code (Ongoing, still some in window.rs)


Miro-Andrin:
My bad hygiene when doing commits has bitten me in the ass. So it's a very messy commit history, but currently the 
code is in a finished state.  

Things that we did:
    
    Moved /generated into /libcraft by moving most of the generated files into feather/core
    and the generators into /libcraft/generators. All the generators have been updated
    such that they pull data from the right folder and output to were I moved the generated
    files. 

    All files in /feather that had 'use generated:{...' now point to an equivalent file 
    in /libcraft. This went mostly without any issues. 
    
    ItemKind and ItemStack had a version in /generated and /libcraft. The ItemKind was identical, but the ItemStack in libcraft had a difference. The libcraft version does not allow for the creation of a stack containing zero items. Therefore every case where feather used the old ItemStack was a bit tricky. 

    My guess is that someone was in the middle of moving the code to the libcraft version, but just stopped for some reason. The function bodies in /feather/common/src/window.rs had to be completely rewritten, and the InventorySlot struct had to fleshed out as well in libcraft.

feather/base/src/anvil/entity.rs
```diff
@@ -236,11 +238,19 @@ pub struct ItemNbt {
 
 impl ItemNbt {
     /// Create an `ItemStack` of the specified item and amount, setting any nbt present.
+    /// Panics if count is zero.
```
ambeeeeee, line 241: I believe the standard for panicking in documentation is like this:

```suggestion
    /// Create an `ItemStack` of the specified item and amount, setting any NBT present.
    ///
    /// # Panics
    /// Panics if `count` is zero.
``` 

At a minimum make sure to add spacing for markdown formatting. 

image

feather/base/src/anvil/entity.rs
```diff
@@ -236,11 +238,19 @@ pub struct ItemNbt {
...
-        ItemStack {
-            count: count as u32,
-            item,
-            damage: nbt.as_ref().map(|n| n.damage).flatten().map(|x| x as u32),
+        match nbt {
+            Some(ItemNbt {
+                damage: Some(damage),
+            }) => ItemStackBuilder::with_item(item)
+                .count(count as u32)
+                .damage((*damage).try_into().unwrap())
```
ambeeeeee, line 248: Is unwrapping here okay? Should we be returning a `Result` or can this never panic?

feather/base/src/anvil/entity.rs
```diff
@@ -236,11 +238,19 @@ pub struct ItemNbt {
...
+                .into(),
+
+            Some(ItemNbt { damage: None }) | None => {
+                ItemStackBuilder::with_item(item).count(count as u32).into()
```
ambeeeeee, line 252: If an `ItemStack` cannot have more than `u8` items, why is the builder using `u32`?

feather/common/src/entities/player.rs
```diff
@@ -17,7 +17,7 @@ pub fn build_default(builder: &mut EntityBuilder) {
 }
 
 /// The hotbar slot a player's cursor is currently on
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Default)]
```
ambeeeeee, line 20: There's discussion for removing the default derive for this structure [here](https://github.com/feather-rs/feather/pull/471#issuecomment-927202657). Please leave your feedback and move forward with whatever we decide.

feather/datapacks/src/recipe.rs
```diff
@@ -340,11 +340,11 @@ impl StonecuttingRecipe {
...
-                damage: None,
-            })
+            Some(
+                ItemStackBuilder::with_item(self.result.into())
+                    .count(self.count.into())
+                    .into(),
+            )
         } else {
             None
         }
```
ambeeeeee, line 350: Yoink all this out into its own branch for later.

feather/datapacks/src/serde_helpers.rs
```diff
@@ -12,24 +17,22 @@ pub struct SerdeItem(Item);
...
         Self {
-            item: s.item.into(),
-            count: s.count,
-            damage: s.damage,
+            item: s.item().into(),
+            count: s.count().try_into().unwrap(),
+            damage: s.damage_taken(),
         }
     }
 }
```
ambeeeeee, line 38: Yoink all this out into its own branch with the other datapack code for later.

feather/datapacks/src/tag.rs
```diff
@@ -301,7 +302,7 @@ impl TagRegistry {
                 name: tag_name.to_string(),
                 entries: item_names
                     .iter()
-                    .map(|e| VarInt(generated::Item::from_name(e.name()).unwrap().id() as i32))
+                    .map(|e| VarInt(Item::from_name(e.name()).unwrap().id() as i32))
                     .collect(),
             });
         }
```
ambeeeeee, line 308: Same as above :(

feather/protocol/src/io.rs
```diff
@@ -554,24 +555,31 @@ impl Readable for Slot {
...
+                ItemStackBuilder::with_item(item)
+                    .count(count)
+                    .apply_damage(tags.map(|t| t.damage).flatten().map(|d| d as u32))
+                    .into(),
+            ))
+            //     ItemStack {
+            //     item,
+            //     count,
+            //     damage: tags.map(|t| t.damage).flatten().map(|d| d as u32),
+            // }
```
ambeeeeee, line 569: Commented out code should be removed.

feather/server/src/systems/player_leave.rs
```diff
@@ -127,14 +127,21 @@ fn create_player_data(
             .to_vec()
             .iter()
             .enumerate()
-            .filter_map(|(slot, maybe_item)| maybe_item.as_ref().map(|item| (slot, item)))
+            .filter(|(_, item)| item.is_filled())
+            .map(|(slot, item)| {
+                if let libcraft_items::InventorySlot::Filled(item) = item {
+                    (slot, item)
+                } else {
+                    unreachable!()
```
ambeeeeee, line 135: Please add a comment stating why this is unreachable.

libcraft/inventory/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "libcraft-inventory"
+version = "0.1.0"
+authors = ["Tracreed <davidalasow@gmail.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+libcraft-items = { path = "../items" }
+parking_lot = "0.11"
```
ambeeeeee, line 11: File is missing a trailing newline ):

libcraft/items/src/item_stack.rs
```diff
@@ -299,3 +342,105 @@ impl Display for ItemStackError {
...
+            damage: None,
+            repair_cost: None,
+            enchantments: vec![],
+        }
+    }
+}
+
+pub struct ItemStackBuilder {
+    item: Item,
+    count: NonZeroU32,
```
ambeeeeee, line 360: Why is this a U32 when count is a `u8`?

ambeeeeee (changes requested):
Overall in the right direction, I know you guys were likely planning to do this but please pull out the tags stuff into its own branch for later.

feather/base/src/anvil/entity.rs
Tracreed: `ItemStack.count` is defined as NonZeroU32. That is why it's being cast to u32 here.

libcraft/items/src/item_stack.rs
Tracreed: count is a u32.

Defman: I think Minecraft uses i32?

feather/common/src/entities/player.rs
Miro-Andrin: We just sidestepped the whole problem, by removing the default implementation. It was only used in one place. We leave the decision to someone else to make.

feather/datapacks/src/tag.rs
Miro-Andrin: File was removed.

feather/base/src/anvil/entity.rs
Miro-Andrin: Nice catch, this is an issue.  I am going to change this.

feather/server/src/systems/player_leave.rs
Miro-Andrin: Done.

libcraft/items/src/item_stack.rs
Miro-Andrin: It might be the case that ItemStacks should allow for negative stack sizes. However, 
i believe this is outside the scope of this refactor.  That was how ItemStack were modelled when we started. This "issue" should be revisited on a separate pull request.

feather/server/src/packet_handlers/inventory.rs
```diff
@@ -54,7 +54,7 @@ pub fn handle_click_window(
     if packet.slot >= 0 {
         client.set_slot(packet.slot, window.item(packet.slot as usize)?.clone());
     }
-    client.set_cursor_slot(window.cursor_item());
+    client.set_cursor_slot(window.cursor_item().clone());
```
Defman, line 57: Safety comment... can we safely do this without introducing a duplication glitch?

feather/server/src/systems/player_leave.rs
Defman: Use filter map instead.

Defman (changes requested):

feather/base/src/anvil/entity.rs
Miro-Andrin: I made damage on ItemStack be a i32, reflecting the fact that vanilla minecraft allows negative damage amounts, making the anvil damage and itemstack damage 
congruent.

feather/server/src/packet_handlers/inventory.rs
Miro-Andrin: Ownership was taken because packets always takes ownership. The value was never stored anywhere so it was safe. Moved the clone closer to the packet creation such that it was made more obvious that it is safe.

Defman (approved):

PR #462, opened 2021-09-11
Health and Fall damage systems

MrGunflame:
# Health and Fall damage systems

## Status

- [ ] Ready 
- [x] Development
- [ ] Hold

## Description

Adds Health and Hunger components. Adds a health and fall damage system.

## Related issues

 #358

## Checklist

### DIFF
--- a/feather/common/src/game.rs
+++ b/feather/common/src/game.rs
@@ -2,10 +2,10 @@ use std::{cell::RefCell, mem, rc::Rc, sync::Arc};
 
 use base::{BlockId, BlockPosition, ChunkPosition, Position, Text, Title};
 use ecs::{
-    Ecs, Entity, EntityBuilder, HasEcs, HasResources, NoSuchEntity, Resources, SysResult,
-    SystemExecutor,
+    Ecs, Entity, EntityBuilder, EntityRef, HasEcs, HasResources, NoSuchEntity, Resources,
+    SysResult, SystemExecutor,
 };
-use quill_common::{entities::Player, entity_init::EntityInit};
+use quill_common::{components::Health, entities::Player, entity_init::EntityInit};
 
 use crate::{
     chat::{ChatKind, ChatMessage},
@@ -228,6 +228,13 @@ impl Game {
     /// necessary block updates.
     pub fn break_block(&mut self, pos: BlockPosition) -> bool {
         self.set_block(pos, BlockId::air())
+    }
+
+    pub fn reset_player(&self, player: EntityRef) {
+        match player.get_mut::<Health>() {
+            Ok(mut health) => *health = Health::new(20),
+            Err(_) => (),
+        }
     }
 }
 
--- a/feather/server/src/client.rs
+++ b/feather/server/src/client.rs
@@ -15,7 +15,9 @@ use common::{
     Window,
 };
 use flume::{Receiver, Sender};
-use packets::server::{Particle, SetSlot, SpawnLivingEntity, UpdateLight, WindowConfirmation};
+use packets::server::{
+    Particle, Respawn, SetSlot, SpawnLivingEntity, UpdateHealth, UpdateLight, WindowConfirmation,
+};
 use protocol::{
     packets::{
         self,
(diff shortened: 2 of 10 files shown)


feather/common/src/game.rs
```diff
@@ -229,6 +229,12 @@ impl Game {
     pub fn break_block(&mut self, pos: BlockPosition) -> bool {
         self.set_block(pos, BlockId::air())
     }
+
+    pub fn reset_player(&self, player: EntityRef) {
```
ambeeeeee, line 233: I think just completely ignoring a failure here is not ideal, maybe return an error of some kind upon failure.

feather/server/src/entities.rs
```diff
@@ -57,3 +58,8 @@ fn spawn_living_entity(entity: &EntityRef, client: &Client) -> SysResult {
     client.send_living_entity(network_id, uuid, pos, kind);
     Ok(())
 }
+
+/// Increasing distance of an entity falling. Used to calculate
+/// fall damage for entities with health.
```
ambeeeeee, line 63: Assuming I understand what this means, I think it would be best to say something similar to 
`/// Distance fallen since an entity was on solid ground.` 
Without the explanation of what the type is for.

MrGunflame:
Should be mostly complete now. Someone glance over it again.

qualterz:
It works quite weird when you bunnyhoping down from low hills, and you also take damage during bunnyhop with some delay.

MrGunflame:
Really hard to reproduce, I believe what happened was that when you hit the ground and jump again before the server ticks, the fall damage is calculated for both jumps.
Resetting the fall distance when moving up should fix this, but this behavior should only be applied to players.

MrGunflame:
That's a really weird one, seems to only happen when, immediately after connecting, you jump from a cliff into another chunk causing the client to reload and for some reason teleporting a bit up, resetting the falling distance. I'm not sure how to solve this in the current implementation while not re-enabling the bunnyhopping issue. I guess there's a reason minecraft uses player velocity to calculate fall damage, but that's not implemented in feather yet (afaik).

Defman:
I think this is mostly do to physics not being implemented on server side, which means we 100% rely on the data from the client.

feather/server/src/client.rs
```diff
@@ -501,6 +503,34 @@ impl Client {
...
+            food: player_hunger.food as i32,
+            food_saturation: player_hunger.saturation as f32,
+        });
+    }
+
+    pub fn respawn_player(&self, gamemode: Gamemode) {
+        let dimension = nbt::Blob::from_reader(&mut Cursor::new(include_bytes!(
+            "../../../assets/dimension.nbt"
+        )))
+        .expect("dimenstion asset is malformed");
```
Nufflee, line 520:
```suggestion
        .expect("dimension asset is malformed");
```

feather/server/src/packet_handlers/health.rs
```diff
@@ -0,0 +1,40 @@
...
+pub fn handle_client_status(
+    game: &mut Game,
+    server: &mut Server,
+    player_id: Entity,
+    packet: ClientStatus,
+) -> SysResult {
+    match packet {
+        ClientStatus::PerformRespawn => {
+            let client_id = game.ecs.get::<ClientId>(player_id).unwrap();
+            let client = server.clients.get(*client_id).unwrap();
```
ambeeeeee, line 16: Can't you `?` these instead of unwrapping?

feather/server/src/packet_handlers/health.rs
```diff
@@ -0,0 +1,40 @@
...
+
+            client.respawn_player(server.options.default_gamemode);
+
+            let player = game.ecs.entity(player_id)?;
+            game.reset_player(player)?;
+
+            let network_id = game.ecs.get::<NetworkId>(player_id).unwrap();
+            let position = game.ecs.get::<Position>(player_id).unwrap();
+            let uuid = game.ecs.get::<Uuid>(player_id).unwrap();
```
ambeeeeee, line 25: Can't you `?` these instead of unwrapping?

ambeeeeee (changes requested):

feather/server/src/packet_handlers/health.rs
MrGunflame: Yea, not sure why I used unwrap there, but returning an error is the better way. Though not sure what error would be appropriate for `server.clients.get` instead of unwrapping if any?

leElvyn:
Is this PR mergable ?

MrGunflame:
Currently this implementation works for mobs, but for players it's a little buggy due to entity physics not being implemented yet. I'd therefore wait until entity physics are implemented before merging this.

PR #464, opened 2021-09-14
GitHub Actions improvements and cross compilation

Defman:
# Improvement to GitHub Actions
Using cross for additional targets
## Status

- [ ] Ready 
- [x] Development
- [ ] Hold

## Checklist


Defman:
Hmm some issue with `bzip2-sys` when compiling for M1. @OxKitsune don't have these issue when compiling on his m1?

Defman:
Should tests be run on all T1 targets on every PR?
Tests currently fail caching :/

Defman:
This PR is still missing publishing, this will be added in the future. Father more there is an issue with test and caching, which is being looked at.

Issue #473, opened 2021-09-18
Download failed for server.jar


Seadragon91:
## Description
When I build feather I got this error:
```
error: failed to run custom build command for `feather-datapacks v0.1.0 (/home/lukas/rust/feather/feather/datapacks)`

Caused by:
  process didn't exit successfully: `/home/lukas/rust/feather/target/debug/build/feather-datapacks-1cf8e9c030678bb2/build-script-build` (exit status: 1)
  --- stderr
  Error: please make sure you have an Internet connection.

  Caused by:
      0: failed to download vanilla server JAR
      1: timed out reading response
warning: build failed, waiting for other jobs to finish...
error: build failed
```

Well my internet works fine. It looks like the function [timeout](https://docs.rs/ureq/2.2.0/ureq/struct.Request.html#method.timeout) also sets the time limit in which the file has to be downloaded. After changing
https://github.com/feather-rs/feather/blob/df3a3a4ae42b0717dd0c50ec3bf3804aed0936c6/feather/datapacks/vanilla_assets/src/lib.rs#L36
the duration to 60, it builds fine. That means the file has to be downloaded in  max 10 seconds.

## Reproduction Steps

1.  Have a slow internet
2. Build feather

## What You Expected to Happen
Successful build.

## Your Environment

- Operating System: Debian
- Operating System Version: Debian GNU/Linux 10 (buster)
- **Feather** Version: latest commit: df3a3a4ae42b0717dd0c50ec3bf3804aed0936c6

Seadragon91:
I guess this can be closed?

PR #474, opened 2021-09-18
Change vanilla JAR download timeout and add warnings

Iaiao:
# Change vanilla JAR download timeout and add warnings

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Description

Changed vanilla JAR download timeout, made it const
Added warnings when the network is very slow (with 2, 4, 6, ... minutes timeout). No one will see this because we don't have a logging frontend in feather-datapacks build script. Also, in build scripts, the stdout is only saved to target/debug/build/feather-datapacks-{hash}/output and not displayed on cargo build's stdout (except for cargo build -vv, which shows too much output to find the warning)

## Related issues

#473 

## Checklist

### DIFF
--- a/feather/datapacks/vanilla_assets/src/lib.rs
+++ b/feather/datapacks/vanilla_assets/src/lib.rs
@@ -1,4 +1,7 @@
 use anyhow::Context;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::SystemTime;
 use std::{
     fs::{self, File},
     io,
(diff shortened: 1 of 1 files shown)


Defman:
Suggestion to switch to reqwest, which have better support for getting chunks. Then advancing the timer on every chunk instead of separate thread.

Iaiao:
reqwest doesn't have chunks in reqwest::blocking, and I'm not sure it's ok to do something like tokio::spawn_blocking just for a simple warning. Also, the warning is for users with bad internet connection, so they may not receive chunks or receive them with a delay, but the warning should not be delayed because of these problems.

PR #483, opened 2021-09-25
Add `ValidatedBlockPosition`

### LINKED ISSUE #476: Change the representation of BlockPosition.
Currently the BlockPosition struct is defined like this:  
```rust
#[repr(C)]
pub struct BlockPosition {
     x: i32,
     y: i32,
     z: i32.
}
```
This is representation has some issue.  First of all the y value can never exceed the build height. And a i32, is overkill for that. It should probably be a i16, allowing for -64 to 4096  that is to be introduced in minecraft 1.17+  -.  However its nice to have a simple representation, so i don't think it's  necessarily a bad idea to keep it like this.  But maybe something to think about. 

Another issue with the current representation , happens when Block Positions are sent to the client.  [wiki.vg](https://wiki.vg/Protocol#Data_types) defines a position like this:
```
> An integer/block position:
    x (-33554432 to 33554431),
    y (-2048 to 2047), 
    z (-33554432 to 33554431) 
    x as a 26-bit integer, followed by y as a 12-bit integer, followed  by z as a 26-bit integer (all signed, two's complement).
```
You can therefore create a BlockPosition strut that can never be sent from the server to the client.  Because the values exceed what is actually written to the client.  When block position is written to the client its done like this:
```rust
impl Writeable for BlockPosition {
    fn write(&self, buffer: &mut Vec<u8>, version: ProtocolVersion) -> anyhow::Result<()> {
        let val = ((self.x as u64 & 0x3FFFFFF) << 38)
            | ((self.z as u64 & 0x3FFFFFF) << 12)
            | (self.y as u64 & 0xFFF);
        val.write(buffer, version)?;

        Ok(())
    }
}
```
This code silently converts from the internal representation to how its sent on the wire. Potentially  sending incoherent positions to the client.   

# Suggested solutions.  
Either change BlockPosition to how wiki.vg specifies it, or create a new position type. Something like ProtocolBlockPosision using wiki.vg's specification, and implement try_into  from BlockPosition to ProtocolBlockPosition, and Into the other way around.

ambeeeeee:
# Add `ValidatedBlockPosition`

## Status

- [X] Ready 
- [ ] Development
- [ ] Hold

## Description

Adds `ValidatedBlockPosition` which checks if a block position is valid in order to increase runtime safety. I've updated all internal code to use this type where applicable, 

## Relevant Issues
Closes #476

## Checklist

- [X] Ran `cargo fmt`, `cargo clippy --all-targets`, `cargo build --release` and `cargo test` and fixed any generated errors!
- [X] Removed unnecessary commented out code
- [X] Used specific traces (if you trace actions please specify the cause i.e. the player)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -753,6 +753,7 @@ dependencies = [
  "arrayvec 0.7.1",
  "bitflags",
  "bitvec",
+ "bytemuck",
  "byteorder",
  "feather-blocks",
  "feather-generated",
--- a/feather/base/Cargo.toml
+++ b/feather/base/Cargo.toml
@@ -32,6 +32,7 @@ smallvec = "1"
 thiserror = "1"
 uuid = { version = "0.8", features = [ "serde" ] }
 vek = "0.14"
+bytemuck = "*"
 
 [dev-dependencies]
 rand = "0.8"
--- a/feather/base/Cargo.toml
+++ b/feather/base/Cargo.toml
@@ -32,7 +32,7 @@ smallvec = "1"
 thiserror = "1"
 uuid = { version = "0.8", features = [ "serde" ] }
 vek = "0.14"
-bytemuck = "*"
+bytemuck = { version = "1", features = ["derive"] }
 
 [dev-dependencies]
 rand = "0.8"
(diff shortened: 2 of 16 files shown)


ambeeeeee:
It should be noted for future development involving `BlockPosition`, Plugins should likely never be allowed to have anything to do with `ValidBlockPosition` because they could maliciously inject invalid data.

PR #492, opened 2021-09-27
Entity sneak and sprint

Tracreed:
# Add Entity Sneaking and Sprinting for other clients.

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Description

Added 2 new systems that send a `EntityMetadata` packet, that tells clients that a certain entity is sneaking or sprinting.
Added pose Enum to EntityMetadata, also changed Const name to better fit.
Added a `send_entity_metadata` to client.

## Checklist

### DIFF
--- a/feather/base/src/metadata.rs
+++ b/feather/base/src/metadata.rs
@@ -20,7 +20,7 @@ pub const META_INDEX_IS_CUSTOM_NAME_VISIBLE: u8 = 3;
 pub const META_INDEX_IS_SILENT: u8 = 4;
 pub const META_INDEX_NO_GRAVITY: u8 = 5;
 
-pub const META_INDEX_ITEM_SLOT: u8 = 6;
+pub const META_INDEX_POSE: u8 = 6;
 
 pub const META_INDEX_FALLING_BLOCK_SPAWN_POSITION: u8 = 7;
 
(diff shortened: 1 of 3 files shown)


Schuwi:
LGTM

Schuwi (approved):

qualterz (approved):

feather/base/src/metadata.rs
```diff
@@ -85,6 +85,30 @@ impl MetaEntry {
...
 
+pub enum Pose {
+    Standing,
+    FallFlying,
+    Sleeping,
+    Swimming,
+    SpinAttack,
+    Sneaking,
+    Dying,
+}
```
ambeeeeee, line 96: Would this make sense to be in `libcraft`? Not saying it would but it might?

feather/server/src/client.rs
```diff
@@ -547,6 +547,16 @@ impl Client {
         });
     }
 
+    pub fn send_entity_metadata(&self, network_id: NetworkId, metadata: EntityMetadata) {
+        if self.network_id().unwrap() == network_id {
```
ambeeeeee, line 551: Will this unwrap ever panic? If not, add a comment asserting why it's safe to unwrap here.

Tracreed: Not calling the function anymore.

ambeeeeee: omg derp I never thought about that or I would have suggested that directly

feather/base/src/metadata.rs
Miro-Andrin: There is a lot of code in base that should be part of libcraft, but that is beyond what this pull request is trying to do.

ambeeeeee: That's not the point, you're adding new code. This new code should be in the most correct place. If you believe it would make sense to keep it there, leave it there. If it would be better placed in `libcraft` put it there.

Tracreed: Pose is only used for metadata as far as I know so it makes sense to keep it with the rest of the metadata code. Just like how the impl ToMetaEntry is defined for all other types that could be expected. It could be argued that all of the metadata code could be in libcraft but the impl and definition of pose should be with the rest of it.

ambeeeeee: It's annoying I can't resolve this comment

ambeeeeee (approved):

Issue #494, opened 2021-09-27
Send entity equipment packet


Miro-Andrin:
# Make it possible to see what items players/mobs have in their main/off-hand

After implementing #492 @Tracreed though about implementing similar feature, but we decided to prioritise working on something else. However if someone is looking for a beginner issue, then this *might* be something to have a look at. 
Note however we are uncertain if this should actually be considered a beginner issue, and would probably classify it as beginner++. 

## Description
Make it so that when mobs or players change their equipment, other clients see the selected item.  So if a player/mob equips a sword or some armour it would be visible to others. 

To do this one has to send the [EntityEquipment](https://github.com/feather-rs/feather/blob/567262c7724d423d3e3ad8dc660b89ef8c72c290/feather/protocol/src/packets/server/play.rs#L948) packet every time an inventory is changed in a visible way. 

To send send a packet to all nearby players, you call the [broadcast_nearby_with](https://github.com/feather-rs/feather/blob/567262c7724d423d3e3ad8dc660b89ef8c72c290/feather/server/src/lib.rs#L125) on the Server struct. 

This code should probably be made into a [system](## "Systems: functions that run each tick. While components are data, systems are logic. They operate on components."). And would therefore be put into [this folder](https://github.com/feather-rs/feather/tree/main/feather/server/src/systems/entity). Remember to register the system like it is done [here](https://github.com/feather-rs/feather/blob/567262c7724d423d3e3ad8dc660b89ef8c72c290/feather/server/src/systems/entity.rs#L15).

I am not sure what the best approach is for detecting a visible change in a inventory slot.
And to do this well the problem might actually turn into an issue that is less beginner friendly. So a first approximation to the problem, would be to trigger the packet every n'th tick, and add a todo comment.  

If you have gotten everything else working and feel like doing something smarter then what I have suggested then I  would ask someone on the discord, or on this issue to see if they agree with your approach.

PR #495, opened 2021-09-28
Entity equipment 

tim-kt:
# Entity equipment

## Status

- [ ] Ready 
- [x] Development
- [ ] Hold

## Description
This PR implements entity equipment packets (i.e. armor, items in the main hand and off-hand).

## Related issues
#494

## Checklist

### DIFF
--- a/feather/server/src/client.rs
+++ b/feather/server/src/client.rs
@@ -27,7 +27,7 @@ use protocol::{
         self,
         server::{
             AddPlayer, Animation, BlockChange, ChatPosition, ChunkData, ChunkDataKind,
-            DestroyEntities, Disconnect, EntityAnimation, EntityHeadLook, JoinGame, KeepAlive,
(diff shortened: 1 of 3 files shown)


tim-kt:
Missing features:
- [ ] Support for survival
- [ ] Send an EntityEquipment packet to players joining

feather/server/src/client.rs
```diff
@@ -558,6 +560,65 @@ impl Client {
...
+            })
+        };
+
+        if equipment.is_empty() {
+            log::warn!("could not get entity equipment");
+        } else {
+            self.send_packet(EntityEquipment {
+                entity_id: network_id.0,
+                entries: equipment,
+            });
```
ambeeeeee, line 618: This would change with the update proposed above.

ambeeeeee (commented):
I know this isn't done, just wanted to suggest a change.

Awesome work so far, I just have one suggestion that should help prevent bugs since there is a requirement on the struct that cannot be enforced at the type level and thus needs to be enforced manually.

ambeeeeee:
Can you write "Closes #494" so github gets the idea about automatically closing the issue?

PR #500, opened 2021-09-30
Added some docs on plugins.

Miro-Andrin:
Started on docs for plugins with help from @Tracreed .

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -1,0 +1,51 @@
+# Feather - Quill - Plugins
+Feather has support for plugins. They can either be run as native code, or through WebAssembly using [Wasmer](https://github.com/wasmerio/wasmer). The API for the two options are the same, but wasmer has a few benefits for the cost of slightly worse performance. ([More details](#Wasm-or-Native)).
+
+## Getting Started
+
+# Building and Running
+To build a quill plugin you will need a tool called [cargo-quill](https://github.com/¨-rs/feather/tree/main/quill/cargo-quill). To install cargo quill you need the [rust build tools](https://www.rust-lang.org/tools/install), and clone Feathers [git repository](https://github.com/feather-rs/feather).
+
+Once that is setup, locate the folder `feather/quill/cargo_quill` in a terminal. Then run the following command.
+
+```bash
+cargo install --path <path to quill>
+```
+This adds `cargo-quill` to path, which can be tested by running.
+```bash
+cargo-quill --help
+```
+
+To create a plugin run the following command.
+```bash
+cargo-quill new <name>
+```
+
+This creates a simple test plugin called <name>. To build this example plugin head into the directory of the plugin (with the Cargo.toml in it) and run:
+
+```bash
+cargo-quill build
+```
+
+
+When building a plugin it can either be done in `release` or `debug` mode. By default it builds in debug mode, but if passed the `--release` flag it builds it in release mode. Debug mode creates slower plugins, however they are built a lot faster, making it quicker to get started. The --native option changs what the plugin is compiled into. By default it compiles to a wasm plugin, but --native gives the native version.
+
+
+Once finished the command will have created a file ending in '.plugin'. It is placed in either `/target/release` or `/target/debug` depending on if the release flag is provided. 
+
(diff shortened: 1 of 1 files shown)


qualterz (approved):

docs/plugins.md
```diff
@@ -0,0 +1,51 @@
+# Feather - Quill - Plugins
+Feather has support for plugins. They can either be run as native code, or through WebAssembly using [Wasmer](https://github.com/wasmerio/wasmer). The API for the two options are the same, but wasmer has a few benefits for the cost of slightly worse performance. ([More details](#Wasm-or-Native)).
+
+## Getting Started
+
+# Building and Running
+To build a quill plugin you will need a tool called [cargo-quill](https://github.com/¨-rs/feather/tree/main/quill/cargo-quill). To install cargo quill you need the [rust build tools](https://www.rust-lang.org/tools/install), and clone Feathers [git repository](https://github.com/feather-rs/feather).
```
tim-kt, line 7:
```suggestion
To build a quill plugin you will need a tool called [cargo-quill](https://github.com/¨-rs/feather/tree/main/quill/cargo-quill). To install cargo quill you need the [rust build tools](https://www.rust-lang.org/tools/install), and clone Feather's [git repository](https://github.com/feather-rs/feather).
```

docs/plugins.md
```diff
@@ -0,0 +1,51 @@
...
+
+Once that is setup, locate the folder `feather/quill/cargo_quill` in a terminal. Then run the following command.
```
tim-kt, line 9:
```suggestion
Once that is setup, locate the folder `feather/quill/cargo-quill` in a terminal. Then run the following command.
```

docs/plugins.md
```diff
@@ -0,0 +1,51 @@
...
+
+```bash
+cargo install --path <path to quill>
```
tim-kt, line 12: What is "<path to quill>"? Do I need to build cargo-quill first?

docs/plugins.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+This creates a simple test plugin called <name>. To build this example plugin head into the directory of the plugin (with the Cargo.toml in it) and run:
+
+```bash
+cargo-quill build
+```
+
+
+When building a plugin it can either be done in `release` or `debug` mode. By default it builds in debug mode, but if passed the `--release` flag it builds it in release mode. Debug mode creates slower plugins, however they are built a lot faster, making it quicker to get started. The --native option changs what the plugin is compiled into. By default it compiles to a wasm plugin, but --native gives the native version.
```
tim-kt, line 31:
```suggestion
By default, the plugin will be built in `debug` mode. The build process will be fast, but the plugin itself will be slower. By specifying the `--release` flag, you can build the plugin in `release` mode, which will take longer, but results in a faster plugin. Without any other flags, the plugin will be compiled to a wasm plugin. If you want to compile it to native code, add the `--native` flag.
```
Different wording

docs/plugins.md
```diff
@@ -0,0 +1,51 @@
...
+
+
+Once finished the command will have created a file ending in '.plugin'. It is placed in either `/target/release` or `/target/debug` depending on if the release flag is provided. 
+
+To test the plugin a feather server binary is needed. Inside the folder where the binary located, you need to locate or create a folder called `plugins`. Put the '.plugin' there and run the server.
```
tim-kt, line 36:
```suggestion
To test the plugin, a feather server binary is needed. Inside the folder where the binary is located, create a folder `plugins` (if it doesn't already exist). Put the '.plugin' there and run the server.
```

tim-kt:
I've submitted some wording changes, nothing too major. What's missing for me as an outsider is what the plugin should look like after I create it with cargo-quill (to see if I made a mistake or not) and what the general structure of a plugin is (`Plugin` trait, implementing this trait, systems, etc).

docs/plugins.md
Miro-Andrin: If i change it to cargo-quill, does that  spark joy?

tim-kt: I'm just not sure why you're using a placeholder. You could say "In the root directory of feather, run the following command" followed by `cargo install --path quill/cargo_quill/` or you could also say "Navigate into quill/cargo_quill and run the following command", followed by `cargo install --path`.

PR #502, opened 2021-10-05
Set more sane default values for light levels

Schuwi:
# Set more sane default values for light levels

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Description

Set block light level to 0 by default. Imo it's better than glowing terrain in the night for now.

## Checklist

- [ ] ~~Used specific traces (if you trace actions please specify the cause i.e. the player)~~


koskja (approved):
Looks good!

PR #503, opened 2021-10-12
Correctly specify packet for interaction (R-click) with an entity

csm-yujinkim:
# Correctly specify packet for interaction (R-click) with an entity

## Status

- [ ] Ready 
- [x] Development
- [ ] Hold

## Description

An incorrect specification for the packet representing interaction (right-click)
from a player to an entity caused the server to crash.

This issue was fixed by changing the data types for the members of the packet representing the target position vectors.

## Checklist

### DIFF
--- a/feather/server/src/connection_worker.rs
+++ b/feather/server/src/connection_worker.rs
@@ -145,7 +145,6 @@ impl Worker {
         tokio::task::spawn(async move {
             let result = reader.race(writer).await.expect("task panicked");
             if let Err(e) = result {
-                log::trace!("Panic error is: {}", e);
                 let message = disconnected_message(e);
                 log::debug!("{} lost connection: {}", username, message);
             }
--- a/feather/server/src/connection_worker.rs
+++ b/feather/server/src/connection_worker.rs
@@ -145,6 +145,7 @@ impl Worker {
         tokio::task::spawn(async move {
             let result = reader.race(writer).await.expect("task panicked");
             if let Err(e) = result {
+                log::trace!("Panic error is: {}", e);
                 let message = disconnected_message(e);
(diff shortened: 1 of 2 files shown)


feather/server/src/connection_worker.rs
```diff
@@ -145,6 +145,7 @@ impl Worker {
         tokio::task::spawn(async move {
             let result = reader.race(writer).await.expect("task panicked");
             if let Err(e) = result {
+                log::trace!("Panic error is: {}", e);
```
Iaiao, line 148: There is no need in this line, `disconnect_message(e)` returns `Debug` error message of the error, which is almost always more useful than `Display`. The exception is `"disconnected"` for `ErrorKind::UnexpectedEof`, but that's more than enough to understand that the player has left the server.

Schuwi: I agree, the error is already logged in line 150, this is superfluous

csm-yujinkim: This was an accidental change I made while I was exploring the repo. I didn't think new changes in my fork were incorporated. I'll be more careful next time.

Schuwi: Ah I see. Then please revert the commit.

axionbuster: That was done.

axionbuster: (I am the same person as @csm-yujinkim) one is my school account and this is my main. Sorry for the switchup but it's because I'm graduating soon.

Schuwi (approved):
Looks good to me, CI should run now

Defman (approved):

PR #506, opened 2021-11-07
Change from `chrono` implementation to impl based on `time` crate

noahcoetsee:
# Change from `chrono` implementation to impl based on `time` crate

## Status

- [ ] Ready 
- [x] Development
- [ ] Hold

## Description

Change from `chrono` implementation to impl based on `time` crate, as it has addressed the issues present in RUSTSEC-2020-0159 in its own security advisory: RUSTSEC-2020-071. This is due to `chrono` no longer being actively maintained.

The new implementation still logs in the exact same time format as the previous `chrono` implementation.

This is necessary to prevent some `cargo audit` issues in the CI/CL Action cycle. There are other issues (involving the same RUSTSEC advisory) currently present due to this issue in crates used by feather such as rust-simple_logger, and I have made a PR there as well to address the issue. Once that PR is (hopefully) merged, we can update the version in `feather-util` and `minecraft-proxy`. The issue is also present in `simple_asn1` and I'm currently working on a PR for that repo as well.

Obviously I'm aware that the _security_ end of this issue isn't super critical given that the library isn't exposed by feather and feather isn't modifying environment variables in many threads (at all), but given it fails the `cargo audit` check in the CI/CL cycle it should be addressed so that future security advisories are not ignored when `cargo audit` is used.

Here's the dependency tree from the `cargo audit`:
image

## Related issues

https://github.com/borntyping/rust-simple_logger/pull/41

## Checklist

### DIFF
--- a/feather/server/src/logging.rs
+++ b/feather/server/src/logging.rs
@@ -19,9 +19,9 @@ pub fn init(level: LevelFilter) {
                 record.module_path().unwrap_or_default()
             };
 
-            let datetime: OffsetDateTime = match OffsetDateTime::now_local().is_ok() {
-                true => OffsetDateTime::now_local().unwrap(),
-                false => OffsetDateTime::now_utc(),
+            let datetime: OffsetDateTime = match OffsetDateTime::now_local() {
+                Ok(x) => x,
+                Err(_) => OffsetDateTime::now_utc(),
             };
             out.finish(format_args!(
                 "{} {:<5} [{}] {}",
--- a/feather/server/src/logging.rs
+++ b/feather/server/src/logging.rs
@@ -1,5 +1,7 @@
 use colored::Colorize;
 use log::{Level, LevelFilter};
+use time::macros::format_description;
+use time::OffsetDateTime;
 
 pub fn init(level: LevelFilter) {
     fern::Dispatch::new()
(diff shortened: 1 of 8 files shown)


feather/server/src/logging.rs
```diff
@@ -16,9 +18,18 @@ pub fn init(level: LevelFilter) {
             } else {
                 record.module_path().unwrap_or_default()
             };
+
+            let datetime: OffsetDateTime = match OffsetDateTime::now_local().is_ok() {
+                true => OffsetDateTime::now_local().unwrap(),
+                false => OffsetDateTime::now_utc(),
+            };
```
panaman67, line 25:
```suggestion
            let datetime: OffsetDateTime = match OffsetDateTime::now_local() {
                Ok(x) => x,
                Err(_) => OffsetDateTime::now_utc(),
            };
```

Iaiao:
You can just remove `simple_logger` from `feather-utils` because it's an unused dependency that we somehow forgot to remove from Cargo.toml. Also, I'm not sure we should use `simple_logger` in `minecraft-proxy` because we use `fern` in `feather-server`, so why not to use the same logging backend in both server and proxy?

`simple_asn1` 0.6 has already moved to `time`, but we need to update it to 0.6 by updating it in https://github.com/caelunshun/rsa-der and updating `rsa` to 0.5.

noahcoetsee:
Okay. I'll remove the unused deps and switch `minecraft_proxy` over to `fern`. I see no reason for it to differ from `feather-server`'s implementation

> but we need to update it to 0.6 by updating it in https://github.com/caelunshun/rsa-der and updating rsa to 0.5.

Is caelunshun still active to update that repo if I make a PR? If he isn't still maintaining that there should be a new fork made under the `feather-rs` organization.

Defman (approved):

Defman:
Huh, seems like we are depending on an old version of the time crate as well... maybe cargo lock?

noahcoetsee:
> Huh, seems like we are depending on an old version of the time crate as well... maybe cargo lock?

Yeah I need to mark this PR as still in progress as I moved on to address the other dependency issues. The old version of time was a dependency of some other dependencies and would generate a similar RUSTSEC segfault error in `cargo audit`.

Defman:
Any updates?

noahcoetsee:
Yeah I've been pretty busy with work and there are some wasm dep. issues that are seemingly irremovable. I'll look into it again in the coming week and see if I can clear it up or not. If not I'll push most of the changes.

dlee13:
Superseded by #512. This can be closed.

PR #512, opened 2022-01-08
Cargo audit fix

koskja:
# Get rid of `cargo-audit` issues

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Description

I have shamelessly stolen @noahcoetsee 's work and finished it to the best of my abilities by running `cargo-update` to bump `wasmer-wasi` to version 2.1.1, which gets rid of the time dependency.

## Related issues

#506 - original PR

## Checklist

### DIFF
--- a/feather/server/Cargo.toml
+++ b/feather/server/Cargo.toml
@@ -17,7 +17,7 @@ ahash = "0.7"
 anyhow = "1"
 base = { path = "../base", package = "feather-base" }
 base64 = "0.13"
-time = { version = "0.3.4", features = ["local-offset", "formatting", "macros"] }
+time = { version = "0.3", features = ["local-offset", "formatting", "macros"] }
 colored = "2"
 common = { path = "../common", package = "feather-common" }
 crossbeam-utils = "0.8"
@@ -37,7 +37,7 @@ plugin-host = { path = "../plugin-host", package = "feather-plugin-host" }
 protocol = { path = "../protocol", package = "feather-protocol" }
 quill-common = { path = "../../quill/common" }
 
-rand = "0.8.0"
+rand = "0.8"
 ring = "0.16"
 
 rsa = "0.5"
--- a/feather/server/Cargo.toml
+++ b/feather/server/Cargo.toml
@@ -17,7 +17,7 @@ ahash = "0.7"
 anyhow = "1"
 base = { path = "../base", package = "feather-base" }
 base64 = "0.13"
-chrono = "0.4"
+time = { version = "0.3.4", features = ["local-offset", "formatting", "macros"] }
 colored = "2"
 common = { path = "../common", package = "feather-common" }
 crossbeam-utils = "0.8"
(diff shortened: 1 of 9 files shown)


noahcoetsee:
Happy to see this finally got done 🤣

feather/server/Cargo.toml
```diff
@@ -36,10 +36,14 @@ parking_lot = "0.11"
...
-rand = "0.7"
+
+rand = "0.8.0"
 ring = "0.16"
-rsa = "0.3"
-rsa-der = "0.2"
+
+rsa = "0.5"
+rsa-der = "0.3"
+base64ct = "=1.1.1"
```
Defman, line 45: Is it necessary to depend on this exact version=

feather/server/src/logging.rs
```diff
@@ -16,9 +18,18 @@ pub fn init(level: LevelFilter) {
...
+            let datetime: OffsetDateTime = match OffsetDateTime::now_local() {
+                Ok(x) => x,
+                Err(_) => OffsetDateTime::now_utc(),
+            };
             out.finish(format_args!(
                 "{} {:<5} [{}] {}",
-                chrono::Local::now().format("%Y-%m-%d %H:%M:%S,%3f"),
+                datetime
+                    .format(format_description!(
+                        "[year]-[month]-[day] [hour]:[minute]:[second],[subsecond digits:3]"
```
Defman, line 30: Can we not just use multiple `{}` and get the year, month, day, hour, minute and second manually? Would it not be faster and safer than using format description?

tools/proxy/Cargo.toml
```diff
@@ -12,5 +12,8 @@ futures-lite = "1"
 argh = "0.1"
 anyhow = "1"
 log = "0.4"
-simple_logger = "1"
 either = "1"
+colored = "2"
+fern = "0.6"
+time = { version = "0.3.4", features = ["local-offset", "formatting", "macros"] }
```
Defman, line 18: Can we depend on `0.3` instead of exactly `0.3.4`?

tools/proxy/src/logging.rs
```diff
@@ -0,0 +1,42 @@
...
+            };
+
+            let datetime: OffsetDateTime = match OffsetDateTime::now_local() {
+                Ok(x) => x,
+                Err(_) => OffsetDateTime::now_utc(),
+            };
+            out.finish(format_args!(
+                "{} {:<5} [{}] {}",
+                datetime
+                    .format(format_description!(
```
Defman, line 29: Same as earlier.

Defman (changes requested):

feather/server/src/logging.rs
dlee13: what's your basis for claiming that would be faster and safer?

koskja: I don't think its safer, but the format has to be recompiled every time the function is called

Defman: I assume its faster to only allocate one String instead of two which gets merged... other than that its just an assumption.

koskja: Solved by @Gator

Defman (approved):

koskja:
Feels done.

Defman (approved):

Schuwi:
Why does this PR activate new features for `wasmer-wasi` and `zip`?

koskja:
Zip actually only gets the `time` feature removed and wasmer-wasi needs these features selected after an update to 2.1.1

PR #513, opened 2022-01-09
Support for custom tablist header and footer

Kecerim24:
# Tablist header and footer

## Status

- [ ] Ready 
- [x] Development
- [ ] Hold

## Description

Created a new `TablistHeaderFooter` that is sent to newly connected players. Other code can update this by changing this structure in `Game::Resources` and causing a `TablistExtrasUpdateEvent` which resends the data to all clients.

## Checklist


feather/server/src/systems/tablist.rs
```diff
@@ -11,11 +11,17 @@ use uuid::Uuid;
 
 use crate::{ClientId, Server};
 
-pub fn register(systems: &mut SystemExecutor<Game>) {
+pub fn register(game: &mut Game, systems: &mut SystemExecutor<Game>) {
+    game.insert_resource(TablistHeaderFooter {
+        header: "{\"text\":\"\"}".to_string(),
+        footer: "{\"text\":\"\"}".to_string(),
+    });
```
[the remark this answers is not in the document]
Kecerim24: I don't think it's that important to have default values, because this isn't a vanilla feature. If anyone needs this, it would be better to use a plugin.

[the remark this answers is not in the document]
Iaiao: Feather is not even close to be fully survival-compatible, so the only usage for Feather in production will be lobbies and mini-games. And there's 99% chance that everyone needs this feature. If someone wants more advanced features (like placeholders, animations), it would still be possible to override the resource by a plugin. I don't know why vanilla server doesn't have this feature, but lack of feature is not a good thing to copy.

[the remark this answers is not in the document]
Kecerim24: Then in what format and where in the config would you like it?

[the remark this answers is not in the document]
Iaiao: > In what format

Just like motd, using `Text::from(str)`. Or there is `libcraft_text::markdown` module that allows writing rich text components in format `@red some text @on_hover @show_text @green some hover text`, but I don't see any existing usage or documentation for this format.

> Where

`[server]` section

feather/server/config.toml
```diff
@@ -13,6 +13,8 @@ motd = "A Feather server"
 max_players = 16
 default_gamemode = "creative"
 view_distance = 12
+default_header = ""
+default_footer = ""
```
Iaiao, line 17: Users won't understand what header/footer is, I think it should be named like `tablist_header` or `playerlist_header`, and have default values to make it even more obvious (like vanilla's "A Minecraft Server" default motd).

feather/common/src/events.rs
```diff
@@ -58,3 +59,12 @@ pub struct ChunkLoadEvent {
 pub struct ChunkLoadFailEvent {
     pub position: ChunkPosition,
 }
+
+#[derive(Debug, Clone)]
+pub struct TablistHeaderFooter {
+    pub header: Text,
+    pub footer: Text,
+}
```
Iaiao, line 67: `feather_common::events` is for events, while `TablistHeaderFooter` is a resource. I'm not sure what file would fit best for this small struct, but definitely not `events.rs`

feather/server/src/systems/tablist.rs
```diff
@@ -63,6 +75,34 @@ fn add_tablist_players(game: &mut Game, server: &mut Server) -> SysResult {
     Ok(())
 }
 
+fn update_tablist_header(game: &mut Game, server: &mut Server) -> SysResult {
```
Iaiao, line 78: It updates not only header, but also footer.

feather/server/src/systems/tablist.rs
```diff
@@ -63,6 +75,34 @@ fn add_tablist_players(game: &mut Game, server: &mut Server) -> SysResult {
...
+            client.send_tablist_header_footer(
+                &header_footer.header.to_string(),
+                &header_footer.footer.to_string(),
+            )
+        });
+    }
+    Ok(())
+}
+
+fn send_tablist_header_on_join(game: &mut Game, server: &mut Server) -> SysResult {
```
Iaiao, line 91: It sends not only header, but also footer.

feather/common/src/events.rs
Kecerim24: Where does it need to be for plugins to get to it?

Iaiao: `quill/common`. Quill doesn't support resources though.

Kecerim24: I'm asking this because I want to know if `TablistHeaderFooter` can be in tablist.rs.

Iaiao: You can create `feather/common/src/tablist.rs` or `quill/common/src/tablist.rs`, but not in `feather/server`. No other crate can access `feather-server`.

Kecerim24:
`quill/common/src/tablist.rs` of course

feather/server/src/options.rs
```diff
@@ -28,6 +28,12 @@ pub struct Options {
     /// The default gamemode for new players.
     pub default_gamemode: Gamemode,
 
+    /// The default tablist header.
+    pub tablist_header: Text,
+
+    /// The default tablist footer.
+    pub tablist_footer: Text,
+
```
Iaiao, line 36: We have `Config` for server config and `Options` for `Server` options (mostly for initial connection handling, status, ping, working with proxies), so I think tablist values should only be in `Config`.

Iaiao (changes requested):

feather/server/src/options.rs
Kecerim24:
```
let server_options = game.resources.get::<Server>().unwrap().options.clone();

game.insert_resource(TablistHeaderFooter {
    header: server_options.tablist_header.clone(),
    footer: server_options.tablist_footer.clone(),
});
```

Then how do I get it from config to use it for `TablistHeaderFooter` in tablist.rs.

Kecerim24: Nevermind, got it.

Iaiao: You can pass it to register(game) function, or even better, make `Config` (or even `Rc<Config>` to make it immutable) a resource and `Options` non-resource. `Options` is never used as a resource (just did a quick search on cs.github)

feather/common/src/events.rs
```diff
@@ -58,3 +58,6 @@ pub struct ChunkLoadEvent {
 pub struct ChunkLoadFailEvent {
     pub position: ChunkPosition,
 }
+
+#[derive(Debug)]
+pub struct TablistExtrasUpdateEvent;
```
Iaiao, line 63: #522 is now merged, so quill-compatible events should be in `quill/common/srs/events/change.rs`, implement Serialize, Deserialize, Clone, be defined in `quill/common/src/component.rs` in `HostComponent` enum and have `bincode_component_impl!` in the end.

Iaiao (changes requested):

Iaiao (approved):

Defman:
I'm not sure if this should be store in the default config. It would be quite easy to write a plugin, which sets the header and footer based on a config file. Going down this rapid hole, would result in a masive config file.

Iaiao:
It's currently impossible to write a plugin that interacts with resources, and we don't have a config API yet. Also, I think it would be a great first experience if we have static tablist header/footer built-in, the same way as we do with motd. Like I said in the previous conversation, everyone wants a pretty tablist in their [mini-game or lobby] server, and it's important to have very-commonly-used features built-in rather than to rely on third-party "must-have" plugins (which we don't even have yet).
btw rust's stdlib sometimes gets replacements for very popular crates (plugins) like `assert_matches` (which is going to land soon) or `lazy_static` (which was superseded by `once_cell` and now we have `#![feature(once_cell)]` in `std::lazy`)

PR #487, opened 2021-09-26
Commands

Iaiao:
# Commands

## Status

- [ ] Ready 
- [x] Development
- [ ] Hold

## Description

This PR adds commands. I've started working on this with lieutenant, but I did't like regex because it's hard to write regexps for something like `@e[advancements={"adv":{criteria:"true"}}]`, it required some refactoring to make compatible structure for Declare Commands packet, `CommandCtx: Send + Sync` wouldn't allow `LifetimelessMut<Game>` and has no concept of tab completion, so I decided to make my own dispatcher instead of rewriting 95% of the existing code.

### Things to do:
- [ ] Vanilla commands 
- - [ ] /help (/?)
- - [ ] /advancement
- - [ ] /attribute
- - [ ] /ban
- - [ ] /ban-ip
- - [ ] /banlist
- - [ ] /bossbar
- - [ ] /clear
- - [ ] /clone
- - [ ] /data
- - [ ] /datapack
- - [ ] /debug
- - [ ] /defaultgamemode
- - [ ] /deop
- - [ ] /difficulty
- - [ ] /effect
- - [ ] /enchant
- - [ ] /execute
- - [ ] /experience (/xp)
- - [ ] /fill
- - [ ] /forceload
- - [ ] /function
- - [ ] /gamemode
- - [ ] /gamerule
- - [ ] /give
- - [ ] /item
- - [ ] /kick
- - [ ] /kill
- - [ ] /list
- - [ ] /locate
- - [ ] /locatebiome
- - [ ] /loot
- - [x] /me
- - [ ] /op
- - [ ] /pardon
- - [ ] /pardon-ip
- - [ ] /particle
- - [ ] /perf
- - [ ] /playsound
- - [ ] /recipe
- - [ ] /reload
- - [ ] /save-all
- - [ ] /save-off
- - [ ] /save-on
- - [ ] /say
- - [ ] /schedule
- - [ ] /scoreboard
- - [ ] /seed
- - [ ] /setblock
- - [ ] /setidletimeout
- - [ ] /setworldspawn
- - [ ] /spawnpoint
- - [ ] /spectate
- - [ ] /spreadplayers
- - [ ] /stop
- - [ ] /stopsound
- - [ ] /summon
- - [ ] /tag
- - [ ] /team
- - [ ] /teammsg (/tm)
- - [ ] /teleport (/tp)
- - [ ] /tell (/msg, /w)
- - [ ] /tellraw
- - [ ] /time
- - [ ] /title
- - [ ] /trigger
[...]

PR #516, opened 2022-01-16
Fixed config seed

ELginas:
# TITLE - Replace

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Description

Removed hardcoded seed in main.rs and replaced it with seed from config file.

## Related issues

_Leave empty if none_

## Checklist


feather/server/src/main.rs
```diff
@@ -64,7 +70,11 @@ fn init_world_source(game: &mut Game, config: &Config) {
     // world otherwise. This is a placeholder:
     // we don't have proper world generation yet.
 
-    let seed = 42; // FIXME: load from the level file
+    let seed = config.world.seed.parse().unwrap_or_else(|_| {
+        let mut hasher = DefaultHasher::new();
```
Defman, line 74: Should use the following hash algorithm `s[0]*31^(n-1) + s[1]*31^(n-2) + … + s[n-1]` which is javas `String.hashCode()`.

Iaiao: Here's my implementation from #487: https://github.com/Iaiao/feather/blob/210fed57a4f54681c35d9d49956b1ced8f78c8fb/feather/server/src/main.rs#L95-L101

ELginas: > Here's my implementation from #487: https://github.com/Iaiao/feather/blob/210fed57a4f54681c35d9d49956b1ced8f78c8fb/feather/server/src/main.rs#L95-L101

Wouldn't it be useless for me to continue if you have already implemented this feature in your own fork?

Defman: Maybe add a few tests, to make sure that the code behave as javas `String.hashCode()`.

Iaiao: > > Here's my implementation from #487: https://github.com/Iaiao/feather/blob/210fed57a4f54681c35d9d49956b1ced8f78c8fb/feather/server/src/main.rs#L95-L101
> 
> Wouldn't it be useless for me to continue if you have already implemented this feature in your own fork?

My implementation doesn't have any tests and won't be merged soon (because it requires all vanilla commands to be implemented), so I think it wouldn't be useless to implement it in a separate pr.

feather/worldgen/src/lib.rs
```diff
@@ -364,6 +364,18 @@ impl BiomeGenerator for StaticBiomeGenerator {
     }
 }
 
+pub fn seed_from_string(s: &str) -> u64 {
```
Defman, line 367: This should maybe be an i64, such that negative seeds also work as expected?

ELginas: Everywhere in feather seeds are defined by u64

Defman: Yes, but it should be parsed as `i64` and then cast to `u64`.

Iaiao: I think it should be `i64` everywhere and only be casted when passing it to RNGs, it would be more convenient for feather or plugin developers who expect `-104019040` seed to be negative, and for end users that expect /seed (or seed value in logs / crash reports) to be the same as in config.toml.

Defman: I agree with the above comment, @ELginas do you want to make this change or should I merge this? We can then create a new issue to change the seed type from `u64` to `i64`.

PR #518, opened 2022-01-17
Change Title's inner value and message type to `Text`

Iaiao:
# Change Title's inner value and message type to `Text`

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Description

This PR changes `Title` to use `Text` instead of json-serialized `String` and messages to use `Text` instead of raw `String`, adds a new quill->host serialization method (json). This change was made because bincode ^1 doesn't support `deserialize_any` but `Text` is an untagged enum and requires `deserialize_any`.

## Checklist

### DIFF
--- a/feather/plugin-host/src/host_calls/entity.rs
+++ b/feather/plugin-host/src/host_calls/entity.rs
@@ -17,12 +17,11 @@ pub fn entity_send_message(
     message_ptr: PluginPtr<u8>,
     message_len: u32,
 ) -> anyhow::Result<()> {
-    let message = cx.read_string(message_ptr, message_len)?;
+    let message = cx.read_json(message_ptr, message_len)?;
     let entity = Entity::from_bits(entity);
-    let _ = cx.game_mut().send_message(
-        entity,
-        ChatMessage::new(ChatKind::System, Text::from(message)),
-    );
+    let _ = cx
+        .game_mut()
+        .send_message(entity, ChatMessage::new(ChatKind::System, message));
     Ok(())
 }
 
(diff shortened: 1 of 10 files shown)


feather/plugin-host/src/host_calls/entity.rs
```diff
@@ -17,12 +17,11 @@ pub fn entity_send_message(
     message_ptr: PluginPtr<u8>,
     message_len: u32,
 ) -> anyhow::Result<()> {
-    let message = cx.read_string(message_ptr, message_len)?;
+    let message = cx.read_json(message_ptr, message_len)?;
```
Defman, line 20: We are blindly trusting plugins to send valid text messages as json without confirming?

Defman: I think this is fine for now, but shows that the current `Text` api is a bit lacking...

Iaiao: The text component is serialized with serde_json in quill/api

quill/api/src/entity.rs
```diff
@@ -92,16 +93,16 @@ impl Entity {
     ///
     /// The message sends as a "system" message.
     /// See [the wiki](https://wiki.vg/Chat) for more details.
-    pub fn send_message(&self, message: impl AsRef<str>) {
-        let message = message.as_ref();
+    pub fn send_message(&self, message: impl Into<Text>) {
+        let message = message.into().to_string();
```
Defman, line 97: We should define some lint which either prefers `to_string()` or `to_owned()`.

Defman (approved):

quill/api/src/entity.rs
Defman: It seems that clippy already lints `"some_string_literal".to_string()` and suggests `to_owned()`.

Iaiao: `.into()` returns `Text` and not `&str`. `to_string()` is derived from `Text`'s `Display` trait implementation, which serializes the text component with `serde_json` and returns the json string.

PR #519, opened 2022-01-17
Add empty world generator as a config option

Iaiao:
# Add empty world generator as a config option

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Checklist


Defman (approved):
Could we maybe rename it to void?

PR #521, opened 2022-01-17
Add an ability to change gamemode from plugins

Iaiao:
# Add an ability to change gamemode from plugins

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Description
Adds GamemodeEvent, InstabreakChangeEvent, FlyingAbilityChangeEvent, BuildingAbilityChangeEvent, and InvulnerabilityChangeEvent. This code was taken from #487 (I had to implement this for /gamemode command), but I think this change deserves a separate PR.

## Checklist

### DIFF
--- a/feather/server/src/client.rs
+++ b/feather/server/src/client.rs
@@ -7,6 +7,7 @@ use std::{
 
 use ahash::AHashSet;
 use flume::{Receiver, Sender};
+use slab::Slab;
 use uuid::Uuid;
 
 use base::{
@@ -20,8 +21,8 @@ use common::{
 use libcraft_items::InventorySlot;
 use packets::server::{Particle, SetSlot, SpawnLivingEntity, UpdateLight, WindowConfirmation};
 use protocol::packets::server::{
-    ChangeGameState, EntityPosition, EntityPositionAndRotation, EntityTeleport, HeldItemChange,
-    PlayerAbilities, StateReason,
+    ChangeGameState, EntityPosition, EntityPositionAndRotation, EntityTeleport, GameStateChange,
+    HeldItemChange, PlayerAbilities,
 };
 use protocol::{
     packets::{
@@ -43,7 +44,6 @@ use crate::{
     network_id_registry::NetworkId,
     Options,
 };
-use slab::Slab;
 
 /// Max number of chunks to send to a client per tick.
 const MAX_CHUNKS_PER_TICK: usize = 10;
@@ -609,8 +609,7 @@ impl Client {
 
     pub fn change_gamemode(&self, gamemode: Gamemode) {
         self.send_packet(ChangeGameState {
-            reason: StateReason::ChangeGameMode,
-            value: gamemode as u8 as f32,
+            state_change: GameStateChange::ChangeGamemode { gamemode },
         })
     }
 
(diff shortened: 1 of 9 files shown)


feather/server/src/systems/gamemode.rs
```diff
@@ -0,0 +1,226 @@
...
+
+use crate::{ClientId, Server};
+use base::anvil::player::PlayerAbilities;
+
+pub fn register(systems: &mut SystemExecutor<Game>) {
+    systems.group::<Server>().add_system(gamemode_change);
+}
+
+fn gamemode_change(game: &mut Game, server: &mut Server) -> SysResult {
+    let mut may_fly_changes = HashMap::new();
```
Defman, line 23: Can be a vector or we could just send the event on change instead?

feather/server/src/systems/gamemode.rs
```diff
@@ -0,0 +1,226 @@
...
+                walk_speed,
+                fly_speed,
+                may_fly: *may_fly,
+                is_flying: *is_flying,
+                may_build: *may_build,
+                instabreak: *instabreak,
+                invulnerable: *invulnerable,
+            });
+    }
+    for (entity, flying) in fly_changes {
```
Defman, line 170: Move up inside the first loop, maybe create a utility function to send event in this case.

Defman (changes requested):

feather/server/src/systems/gamemode.rs
```diff
@@ -0,0 +1,226 @@
...
+            mut instabreak,
+            mut may_build,
+            mut invulnerable,
+            gamemode,
+            prev_gamemode,
+        ),
+    ) in game
+        .ecs
+        .query::<(
+            &GamemodeEvent,
```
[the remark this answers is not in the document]
Iaiao: I wanted to name it like this (and I did in the commands PR), but we have `CreativeFlyingEvent` and not `CreativeFlyingChangeEvent`, `SprintEvent` instead of `SprintToggleEvent`, `SneakEvent`, so I changed it to be in the same style. Forgot to change `InstabreakChangeEvent` though.

feather/server/src/systems/gamemode.rs
Iaiao: If I remember correctly, I was having trouble with borrow checker because `game` is already mutably borrowed because of query. It can't be a vector because its value may be true or false or nothing (property not changed), and we have to dispatch an event in the first two cases, but not the third.

Defman: If you extract ecs into a separate var the you should be able to do it. Also a vector of triples should be faster than a hashmap, if you can't get around the borrow checker.

feather/server/src/client.rs
```diff
@@ -602,6 +607,13 @@ impl Client {
         self.send_packet(HeldItemChange { slot });
     }
 
+    pub fn change_gamemode(&self, gamemode: Gamemode) {
+        self.send_packet(ChangeGameState {
+            reason: StateReason::ChangeGameMode,
+            value: gamemode as u8 as f32,
```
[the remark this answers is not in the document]
Iaiao: ChangeGameState requires it to be f64, so I did it in a different way.

feather/server/src/systems/gamemode.rs
```diff
@@ -0,0 +1,200 @@
...
+                fly_speed,
+                may_fly: *may_fly,
+                is_flying: *is_flying,
+                may_build: *may_build,
+                instabreak: *instabreak,
+                invulnerable: *invulnerable,
+            });
+    }
+    for (entity, flying) in fly_changes {
+        if flying {
```
Defman, line 169: This if statement is redundant, just create `CreativeFlyingEvent` with the value of `flying`?

Defman (approved):

PR #522, opened 2022-01-18
Add events to quill API

Iaiao:
# Add events to quill API

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Description

This PR changes quill API to make `entity_set_component` host call accept a `variant: u8` param, 0 = insert component, 1 = insert entity event, 2 = insert event. Also adds `quill::Game::insert_event` and `quill::Entity::insert_event`

## Checklist

### DIFF
--- a/feather/plugin-host/src/host_calls/component.rs
+++ b/feather/plugin-host/src/host_calls/component.rs
@@ -48,6 +48,13 @@ struct SetComponentVisitor<'a> {
     entity: Entity,
     bytes_ptr: PluginPtr<u8>,
     bytes_len: u32,
+    variant: SetComponentVariant,
+}
+
+enum SetComponentVariant {
+    SetComponent,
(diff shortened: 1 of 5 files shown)


feather/plugin-host/src/host_calls/component.rs
```diff
@@ -76,6 +87,7 @@ pub fn entity_set_component(
     component: u32,
     bytes_ptr: PluginPtr<u8>,
     bytes_len: u32,
+    variant: u8,
```
Defman, line 90: Might be better to expose some host calls for event tracker instead. Such that it is possible to model this as two host calls. Ie adding a component and adding the component to the event tracker.

Defman (changes requested):

Defman (approved):

PR #527, opened 2022-01-21
Libcraft item refactoring

koskja:
# Libcraft Item

## Status

- [x] Ready 
- [ ] Development
- [ ] Hold

## Description

Improved code quality in the `libcraft_items` crate.

## Checklist


libcraft/items/src/inventory_slot.rs
```diff
@@ -13,33 +13,37 @@ pub enum InventorySlot {
...
 impl InventorySlot {
+    /// Creates a new instance with the type `kind` and `count` items
+    #[must_use]
+    #[allow(clippy::missing_panics_doc)]
     pub fn new(kind: Item, count: u32) -> Self {
         if count == 0 {
             Self::Empty
         } else {
-            Self::Filled(ItemStack::new(kind, count).unwrap())
+            Self::Filled(ItemStack::new(kind, count).expect("`count` cannot equal zero"))
```
Iaiao, line 37: `count` cannot be 0 here.

Iaiao (changes requested):

libcraft/items/src/inventory_slot.rs
Defman: Could use `ItemStack::new(kind, count).map(Self::Filed).unwrap_or_default()`.

libcraft/items/src/inventory_slot.rs
```diff
@@ -13,33 +13,35 @@ pub enum InventorySlot {
...
+        Self::Empty
     }
 }
 
 impl From<Option<ItemStack>> for InventorySlot {
     fn from(it: Option<ItemStack>) -> Self {
         match it {
-            Some(item) => InventorySlot::Filled(item),
-            None => InventorySlot::Empty,
+            Some(item) => Self::Filled(item),
```
Defman, line 23: Can replace this with a simple map and unwrap default. ie `it.map(Self::Filled).unwrap_or_default()`.

libcraft/items/src/inventory_slot.rs
```diff
@@ -99,31 +104,34 @@ impl InventorySlot {
     }
 
     /// Returns the number of items stored in the inventory slot.
-    pub fn count(&self) -> u32 {
+    #[must_use]
+    pub const fn count(&self) -> u32 {
         match self {
-            InventorySlot::Filled(stack) => stack.count(),
-            InventorySlot::Empty => 0,
+            Self::Filled(stack) => stack.count(),
```
Defman, line 110: Maybe implement `deref` into an `Option` and then use `map` and `unwrap_or_default`?

libcraft/items/src/item_stack.rs
```diff
@@ -94,62 +92,72 @@ impl ItemStack {
...
         self.count.get()
     }
 
-    /// Removes some items from this `ItemStack`.
+    /// Removes some items from this `ItemStack` and return the new item count.
+    /// # Errors
+    /// Returns `NotEnoughItems` if the new stack would be empty.
+    #[allow(clippy::missing_panics_doc)]
     pub fn remove(&mut self, count: u32) -> Result<u32, ItemStackError> {
         if self.count.get() <= count {
```
Defman, line 150: Maybe use checked subtraction.

Defman (changes requested):

libcraft/items/src/inventory_slot.rs
koskja: I think that would unneccesarily reduce code readability for little to no reason.

libcraft/items/src/item_stack.rs
koskja: `NonZeroU32` doesn't implement `checked_sub`, can you explain what you meant?

libcraft/items/src/inventory_slot.rs
Defman: readability? `self.map(ItemStack::count).unwrap_or_default()`. Pretty readable in my opinion, take the slot and map it to the itemstack count or return default. Instead of default you could just write `unwrap_or(0)`.

libcraft/items/src/item_stack.rs
Defman: Instead of check that `count.get() >= count` and then doing the subtraction. Ie
```rust
self.count = self
  .count
  .get()
  .checked_sub(count)
  .and_then(NonZeroU32::new)
  .ok_or(ItemStackError::NotEnoughItems)?;
Ok(self.count.get())
```
Roughly

FussballAndy: But shouldn't an itemstack be removed when we remove exactly the amount of items it has? With this code we would then fail instead of delete the itemstack

koskja: The thing is, an itemstack cannot have 0 items and I don't think having two separate errors is neccessary.

koskja:
Anything else?

libcraft/items/src/inventory_slot.rs
```diff
@@ -99,31 +98,31 @@ impl InventorySlot {
...
-    /// such that the new could is not greater then self.stack_size().
+    /// such that the new could is not greater then `self.stack_size`().
     /// And that the slot actually contains an item.   
     fn add_count(&mut self, n: u32) {
-        match self {
-            InventorySlot::Filled(x) => x.add(n).unwrap(),
-            InventorySlot::Empty => panic!("add count called on empty inventory slot!"),
-        };
+        self.option_mut()
+            .expect("add count called on empty inventory slot!")
```
Defman, line 111: I don't like this panic. Adding to count of an empty slot is still an empty slot?

Defman (changes requested):
I don't think this api should ever panic.

libcraft/items/src/inventory_slot.rs
koskja: It is an internal function, so I think aborting on any logic error is ok.

